PR #4, opened 2017-09-28
Run this addon from a Docker Script Environment

bamedro:
Run this addon from a Docker Script Environment so it can run on any machine which is capable of running a Docker container.

### DIFF
--- a/Addons/resources/catalog/Gantt_chart.xml
+++ b/Addons/resources/catalog/Gantt_chart.xml
@@ -5,7 +5,9 @@
      xsi:schemaLocation="urn:proactive:jobdescriptor:3.8 http://www.activeeon.com/public_content/schemas/proactive/jobdescriptor/3.8/schedulerjob.xsd"
     name="Gantt_chart" projectName="Scheduler Analytics"
     priority="normal"
-    onTaskError="continueJobExecution">
+    onTaskError="continueJobExecution"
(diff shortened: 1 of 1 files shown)


Addons/resources/catalog/Gantt_chart.xml
```diff
@@ -28,22 +30,37 @@
...
+paHomeContainer = variables.get("PA_SCHEDULER_HOME") 
+proActiveHomeVolume = '-v '+paHomeHost +':'+paHomeContainer+' ' 
+# Prepare working directory (For Dataspaces and serialized task file) 
+workspaceHost = localspace 
+workspaceContainer = localspace 
+workspaceVolume = '-v '+localspace +':'+localspace+' ' 
+# Prepare container working directory 
+containerWorkingDirectory = '-w '+workspaceContainer+' ' 
+# Save pre execution command into magic variable 'preJavaHomeCmd', which is picked up by the node 
+preJavaHomeCmd = dockerRunCommand + dockerParameters + proActiveHomeVolume + workspaceVolume + containerWorkingDirectory + containerName
```
paraita: what kind of magic is that ?

bamedro: A standard Docker Script Environment. I only changed the Dcoker image

PR #21, opened 2017-11-20
 rename bucket's names to follow naming convention

ShatalovYaroslav:
delete ordered_bucket_list as it is no need for it any more


ShatalovYaroslav:
We will still use the ordered_bucket_list

PR #45, opened 2018-01-23
Adding icons to basic-workflows

marcocast:


Cron/resources/catalog/start_every_2_min.xml
```diff
@@ -13,6 +13,10 @@
 			<description>
         <![CDATA[ The simplest task, ran by a javascript engine. ]]>
       </description>
+                  <genericInformation>
+	<info name="task.icon"
```
luispinedamx: I guess the indentation is rendered automatically here. But the rest of tasks look so clean with the proper indentation that this one got my attention.

marcocast: yes u were right

PR #52, opened 2018-02-08
Calendars

marcocast:


Calendars/resources/catalog/EveryMonday.json
```diff
@@ -0,0 +1,12 @@
+{"description":"Every Monday at 9:00 AM excluding holidays","cron":"0 0 9 ? * MON *","exclusion_calendars":[{
+				"calendar": {"url": "https://www.officeholidays.com/ics/ics_country_code.php?iso=US"},
```
chloegugli: The exclusions/inclusions I have have a "catalog" object in "calendar": {"calendar" : { "catalog": null, "url":"..." } , "rule" : { "action" : "..." } }.
It's never used, always null, it's not possible to set it with the job planner portal. Should I remove it or you add it?

PR #60, opened 2018-02-13
Added cognitive services

marcocast:


AzureCognitiveServices/METADATA.json
```diff
@@ -0,0 +1,325 @@
...
+				"file" : "resources/catalog/Custom_Decision_Service.xml"
+			},
+			
+			
+			
+			
+			
+			
+			
+
```
ShatalovYaroslav: empty space here is not needed

build.gradle
```diff
@@ -1,3 +1,6 @@
+import groovy.io.FileType
```
ShatalovYaroslav: why we need it?

PR #93, opened 2018-03-07
add OrderedBucketListTest

ShatalovYaroslav:


src/test/java/org/ow2/proactive/OrderedBucketListTest.java
```diff
@@ -0,0 +1,33 @@
...
+
+    @Test
+    public void testOrderedBucketList() throws IOException {
+
+        List<String> fileLinesList = Files.readAllLines(Paths.get(ORDERED_BUCKET_LIST_FILE_NAME));
+
+        assertThat("The ordered bucket list file should present in project: " + ORDERED_BUCKET_LIST_FILE_NAME,
+                   fileLinesList,
+                   notNullValue());
+
```
Mykhailenko: Are you sure Files.readAllLines will return null if there is nothing?

src/test/java/org/ow2/proactive/OrderedBucketListTest.java
```diff
@@ -0,0 +1,33 @@
...
+        assertThat("The ordered bucket list file should contain only 1 line: ", fileLinesList.size(), is(1));
+
+        assertThat("The ordered bucket list should not contain spaces: ",
+                   fileLinesList.get(0).contains(" "),
+                   is(false));
```
Mykhailenko: some clarification here would be helpfull. why do you take only first line. Maybe fileLinesList is empty..

src/test/java/org/ow2/proactive/OrderedBucketListTest.java
ShatalovYaroslav: you are right, the method will throw the exception. So no need for this check

src/test/java/org/ow2/proactive/OrderedBucketListTest.java
ShatalovYaroslav: It should be only 1 line. Exactly this it's checked in previous assertion

Mykhailenko: ok right

PR #104, opened 2018-03-27
Add REAL first version of sftp-connector task

amouhoub:
THIS IS THE REAL FIRST COMMIT OF THE NEW VERSION of the sftp-connector workflow


DataConnectors/resources/catalog/SFTP.xml
```diff
@@ -8,36 +8,52 @@
     onTaskError="continueJobExecution"
      maxNumberOfExecution="2">
   <variables>
-    <variable name="FTP_HOSTNAME" value="10.220.248.146" />
-    <variable name="INPUT_PATH" value="/home/activeeon/Bureau/folder/" />
-    <variable name="OUTPUT_PATH" value="destination" />
-    <variable name="FTP_PORT" value="21" />
-    <variable name="EXTRACT_ARCHIVE" value="True" model="PA:Boolean"/>
+    <variable name="SFTP_HOSTNAME" value="localhost" />
+    <variable name="SFTP_PORT" value="22" />
```
jrochas: The `PA:Integer` model does not exist? Or we cannot apply it for the port?

amouhoub: Yes, it does exist and it works. In fact, we thought we were using it but I think we lost this sometime in the process... Thank you for the reminder ;)

PR #115, opened 2018-04-11
Sql connectors

Aminelouati:
In this PR we: 
1- Add the Documentation field in task GI 
2-Update java home in fork env
3-Correct postgres connector workflow
4-Remove useless post script in some workflows

### DIFF
--- a/DataConnectors/resources/catalog/cassandra_connector.xml
+++ b/DataConnectors/resources/catalog/cassandra_connector.xml
@@ -44,6 +44,7 @@ The imported data is exported in a JSON format using the variable $DATAFRAME_JSO
           <script>
             <code language="python">
               <![CDATA[
+# In the Java Home location field, use the value: "/usr" to force using the JRE provided in the docker image below (Recommended).
 #Be aware, that the prefix command is internally split by spaces. So paths with spaces won't work.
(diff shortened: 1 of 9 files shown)


DataConnectors/resources/catalog/cassandra_connector.xml
```diff
@@ -37,8 +37,9 @@ The imported data is exported in a JSON format using the variable $DATAFRAME_JSO
       </variables>
       <genericInformation>
         <info name="task.icon" value="/automation-dashboard/styles/patterns/img/wf-icons/cassandra.png"/>
+        <info name="Documentation" value="https://www.activeeon.com/public_content/documentation/latest/user/ProActiveUserGuide.html#_nosql"/>
       </genericInformation>
-      <forkEnvironment javaHome="${PA_SCHEDULER_HOME}/jre" >
+      <forkEnvironment javaHome="/usr" >
```
fviale: Maybe add an xml comment which explains why javaHome is configured this way.

I understand that this is because java is preinstalled in the docker container and is accessible via.
/usr/bin/java, but it is not obvious

Aminelouati: Ok. It will be added.

PR #116, opened 2018-04-11
add a new deep learning workflow for fake celebrity faces generation

ImenBizid:


DeepLearningWorkflows/resources/catalog/Fake_Celebritiy_Faces_Generation.xml
```diff
@@ -0,0 +1,552 @@
...
+      <description>
+        <![CDATA[ This task defines some input, here strings to be processed. ]]>
+      </description>
+      <variables>
+        <variable name="NB_IMAGES" value="64" inherited="false" model="PA:Integer"/>
+      </variables>
+      <genericInformation>
+        <info name="TASK.ICON" value="/automation-dashboard/styles/patterns/img/wf-icons/controls_replicate.png"/>
+      </genericInformation>
+      <forkEnvironment javaHome="${PA_SCHEDULER_HOME}/jre" >
```
Aminelouati: Isn't rather <forkEnvironment javaHome="/usr" > ?

ImenBizid: We can also use javaHome="${PA_SCHEDULER_HOME}/jre".
javaHome="${PA_SCHEDULER_HOME}/jre" is used for all machine learning tasks  and javaHome="/usr" is used for all deep learning tasks. But, I think that it would be better to use javaHome="/usr"  as you suggested for all the tasks in order to not have to install java in the docker image. I will change it

PR #117, opened 2018-04-11
Add Greenplum Database connector

amouhoub:
In this PR, we add a connector to import data from Greenplum Database


amouhoub:
@marcocast Fixed

PR #130, opened 2018-05-02
add Azure Data Lake connector

luispinedamx:
- Added the first version Azure Data Lake connector.
- Tested locally and it works fine.
- Some features are subject to improvement after review: display result, connection, support for Windows.
- Pending to check if the Documentation link is correct (waiting for it to be available on trydev).


DataConnectors/resources/catalog/Azure_Data_Lake.xml
```diff
@@ -0,0 +1,236 @@
...
+            <![CDATA[
+#Verify installation of Azure CLI
+if ! type "az" >/dev/null 2>&1; then
+  echo "Installing Azure CLI"
+
+  # Update sources and install key
+  AZ_REPO=$(lsb_release -cs)
+  echo "deb [arch=amd64] https://packages.microsoft.com/repos/azure-cli/ $AZ_REPO main" | \
+     sudo tee /etc/apt/sources.list.d/azure-cli.list
+  sudo apt-key adv --keyserver packages.microsoft.com --recv-keys 52E16F86FEE04B979B07E28DB02C46DF417A0893
```
Aminelouati: Is it normal to have the key here and not as a third party credential ?

DataConnectors/resources/catalog/Azure_Data_Lake.xml
```diff
@@ -0,0 +1,236 @@
...
```
Aminelouati: Same comment here for the key

DataConnectors/resources/catalog/Azure_Data_Lake.xml
```diff
@@ -0,0 +1,236 @@
...
+    <variable name="AZ_DLA_OUTPUT" value="my_output_file.csv" />
+    <variable name="AZ_DLA_SCRIPT" value="usql.txt" />
+    <variable name="AZ_DLS_ACCOUNT" value="my_dls_account" />
+  </variables>
+  <description>
+    <![CDATA[ This workflow allows to submit a U-SQL job to Azure Data Analytics and retrieve the result file. ]]>
+  </description>
+  <genericInformation>
+    <info name="bucketName" value="data-connectors"/>
+    <info name="Documentation" value="https://www.activeeon.com/public_content/documentation/latest/user/ProActiveUserGuide.html#_azure_data_lake"/>
```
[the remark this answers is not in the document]
luispinedamx: thanks! I think I just copy-pasted it from an outdated source.

DataConnectors/resources/catalog/Azure_Data_Lake.xml
luispinedamx: This is a Microsoft generic key, to be effective as of May, it is not related to the user's credentials.

DataConnectors/resources/catalog/Azure_Data_Lake.xml
```diff
@@ -0,0 +1,236 @@
...
+    onTaskError="continueJobExecution"
+     maxNumberOfExecution="2"
>
+  <variables>
+    <variable name="AZ_CRON_MONITOR" value="* * * * *" />
+    <variable name="AZ_DLA_ACCOUNT" value="my_dla_account" />
+    <variable name="AZ_DLA_JOB" value="my_dla_job" />
+    <variable name="AZ_DLA_OUTPUT" value="my_output_file.csv" />
+    <variable name="AZ_DLA_SCRIPT" value="usql.txt" />
+    <variable name="AZ_DLS_ACCOUNT" value="my_dls_account" />
```
amouhoub: I would advise against using mockup values as default values because the user might think them as defaults or might forget to notice that they are not filled yet.
I think it is better to use blank values for `AZ_DLA_ACCOUNT, AZ_DLA_JOB , AZ_DLA_OUTPUT,  AZ_DLA_SCRIPT , AZ_DLS_ACCOUNT `
Furthermore, you can perform a check if the mandatory variables are filled and display an error message in the logs stating that `X` cannot be empty or something like that (chack what we did for other connectors in python)

DataConnectors/resources/catalog/Azure_Data_Lake.xml
amouhoub: if you want, you can leave a comment above to clarify that for users

DataConnectors/resources/dataspace/usql.txt
```diff
@@ -0,0 +1,13 @@
+@a  =
```
amouhoub: I think usql scripts have an `*.usql` extension like in:
https://docs.microsoft.com/fr-fr/azure/data-lake-analytics/data-lake-analytics-data-lake-tools-get-started
It's better to rename the file with a meaningful name and change the extension.

luispinedamx: you are absolutely right :)

PR #134, opened 2018-05-03
Zookeeper, Kafka, Storm, Visdom and Dataflow updated with respect to PCA

nebilbenmabrouk:
Zookeeper, Kafka, Storm, Visdom and Dataflow updated with respect to PCA
- New client jar 'cloud-automation-service-client-8.1.0-SNAPSHOT.jar' used to call cloud-automation-service
- METADTA.json updated
- Workflow files updated


DataStreaming/resources/catalog/Kafka_Storm_Visdom_Dataflow.xml
```diff
@@ -1,557 +1,831 @@
...
+println(session_id)
+
+// Define PCA URL
+def scheduler_rest_url = variables.get("PA_SCHEDULER_REST_URL")
+def pca_url = scheduler_rest_url.replaceAll("/rest\\z", "/cloud-automation-service")
+
+// Connect to APIs
+def api_client = new ApiClient()
+api_client.setBasePath(pca_url)
+api_client.setDebugging(true)
```
mbenguig: could be set to false or put in comment

DataStreaming/resources/catalog/Kafka_Storm_Visdom_Dataflow.xml
```diff
@@ -1,557 +1,831 @@
...
+<?xml version="1.0" encoding="UTF-8"?>
+<job
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xmlns="urn:proactive:jobdescriptor:3.10"
+     xsi:schemaLocation="urn:proactive:jobdescriptor:3.10 http://www.activeeon.com/public_content/schemas/proactive/jobdescriptor/3.10/schedulerjob.xsd"
+    name="Kafka_Storm_Visdom_Dataflow" projectName="Data Stream Processing"
+    priority="normal"
+    onTaskError="pauseJob"
+     maxNumberOfExecution="2"
>
```
mbenguig: could you just confirm the workflow is still valid (from the studio) with ^M.. otherwise just remove it

Kafka/resources/catalog/Kafka_Delete.xml
```diff
@@ -0,0 +1,120 @@
...
+          <code language="groovy">
+            <![CDATA[
+// Acquire service instance id
+def service_instance_id = Long.valueOf(variables.get("PCA_INSTANCE_ID"))
+
+// Construct synchro channel name
+channel = "Service_Instance_" + service_instance_id  
+
+//Acquire docker container to delete from synchro channel
+def instance_name =  synchronizationapi.get(channel,"INSTANCE_NAME")
```
fviale: are you sure there is no need to call createChannelIfAbsent here ?

Storm/resources/catalog/Storm.xml
```diff
@@ -1,304 +1,504 @@
...
+api_client.setDebugging(true)
+def service_instance_rest_api = new ServiceInstanceRestApi(api_client)
+
+// Acquire service instance id
+def service_instance_id = Long.valueOf(variables.get("PCA_INSTANCE_ID"))
+println(service_instance_id)
+
+// Construct synchro channel
+channel = "Service_Instance_" + service_instance_id  
+//synchronizationapi.createChannelIfAbsent(channel, false)
```
fviale: using the channel without making sure it is not previously created is risky.

DataStreaming/resources/catalog/Kafka_Storm_Visdom_Dataflow.xml
nebilbenmabrouk: done

Kafka/resources/catalog/Kafka_Delete.xml
nebilbenmabrouk: yes, each platform (zookeeper, kafka, storm, visdom) uses a channel to sunchronize between activation and delete workflows, and also between dependant workflows (like kafka and zookeeper).

Kafka/METADATA.json
```diff
@@ -7,6 +7,10 @@
 		"tags": ["Cloud Automation", "Kafka", "Big Data", "Data Streaming"],
 		"version": "1.0"
 	},
+	"dataspace" : {
+		"target" : "global",
+		"files" : ["resources/dataspace/cloud-automation-service-client-8.1.0-SNAPSHOT.jar"]
```
amouhoub: @nebilbenmabrouk  Just a question, do you really need to provide the PCA jar for the workflow and include it in the dataspace?
Isn't it possible to use this directly if it's packaged with Proactive?

nebilbenmabrouk: It would be better to separate PA dependencies from those of workflows

amouhoub: but if it's already packaged with proactive anyway, why duplicate a jar and move back&forth between PA nodes and the data space?

nebilbenmabrouk: It is packaged with the data (and not the dependencies), and it is not duplicated, we will have a single jar in the global dataspace. I agree with you for the 'move back&forth between PA nodes and the data space', but the PCA client should be separated from PA, and anyway  I cannot decide about including the client jar in PA dependencies, this is related to the cloud-automation-service project

amouhoub: @nebilbenmabrouk understood, thank you for your answer

PR #149, opened 2018-05-11
Update SAP connector to match naming convention

luispinedamx:


ImenBizid:
Hi Luis,
You also need to add a task GI task.documentation

PR #167, opened 2018-05-23
Change preview of Azure DLA connector

luispinedamx:


DataConnectors/resources/catalog/Azure_Data_Lake.xml
```diff
@@ -241,8 +252,56 @@ connection_string.execute()
...
 println(download_string.execute().text)
 //Display file
-println("Output file " + variables.get("AZ_DLA_OUTPUT") + ":")
-println(["cat", variables.get("AZ_DLA_OUTPUT")].execute().text)
+println("Raw output file: " + variables.get("AZ_DLA_OUTPUT"))
+println("You can visualize the formatted output in the Preview tab")
+
+def reader = new CSVReader(new FileReader(new File(variables.get("AZ_DLA_OUTPUT"))))
+def writer = new StringWriter()
+def builder = new groovy.xml.MarkupBuilder(writer)
```
Aminelouati: I would prefer that u add workflow variable having two values: CSV, HTML to allows the user to choose the type of the output. U can also push this csv file into the global space using data management.

luispinedamx: The CSV file comes from Azure directly, and is actually pushed to the user space. The HTML is just an extra user-friendly option to visualize the same data. As usual, it is optional to download the HTML.

Aminelouati: ok ;)

PR #170, opened 2018-05-24
Update s3

amouhoub:
In this PR, we update the name of the S3 connector workflow
Note that we don't change the file name


Aminelouati:
re-test this please

PR #187, opened 2018-06-04
Add node source examples

jrochas:
- add node source json examples in the catalog, under the kind NodeSource

### DIFF
--- a/NodeSources/METADATA.json
+++ b/NodeSources/METADATA.json
@@ -1,0 +1,114 @@
+{
+	"metadata": {
+		"slug": "nodesources",
+		"name": "Node Sources",
+		"short_description": "Node Sources Definitions",
+		"author": "ActiveEon's Team",
+		"tags": ["Node Sources", "Node Sources Definitions"],
+		"version": "1.0"
+	},
+	"catalog" : {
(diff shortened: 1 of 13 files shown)


NodeSources/METADATA.json
```diff
@@ -0,0 +1,114 @@
...
+					"kind": "nodesource",
+					"commitMessage": "First commit",
+					"contentType": "application/json"
+				},
+				"file" : "resources/catalog/OpenStack.json"
+			},
+			{
+				"name" : "VMware",
+				"metadata" : {
+					"kind": "nodesource",
```
mbenguig: (nodesource or NodeSource)

jrochas: I have confirmation from Chloé and that the kind is lowercased in any case.

ShatalovYaroslav: @jrochas I give the confirmation that in catalog service and catalog portal the kind can be searched case-insensitive. But to be in line with naming for all other kinds that we have in official list from our managers, please put "NodeSource".

jrochas: @ShatalovYaroslav ok, I see your point, no problem I will change it.

PR #188, opened 2018-06-04
Changing kinds of worklfows

chloegugli:


chloegugli:
@amouhoub Catalog is not case sensitive so it doesn't matter. I put like that because that's how it's displayed in the portal

PR #217, opened 2018-07-16
add Cassandra PCA service and client-side workflows

Aminelouati:
In this PR, we:
1/ Add a new Cassandra PCA service composed of three states (Started, Stopped and Finished)
2/ Add two workflows to the database-services bucket showing how to use this service from the client-side

### DIFF
--- a/Cassandra/resources/catalog/Cassandra.xml
+++ b/Cassandra/resources/catalog/Cassandra.xml
@@ -1,0 +1,252 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<job
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xmlns="urn:proactive:jobdescriptor:3.10"
+     xsi:schemaLocation="urn:proactive:jobdescriptor:3.10 http://www.activeeon.com/public_content/schemas/proactive/jobdescriptor/3.10/schedulerjob.xsd"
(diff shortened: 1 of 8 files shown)


Cassandra/resources/catalog/Cassandra.xml
```diff
@@ -0,0 +1,252 @@
...
+<job
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xmlns="urn:proactive:jobdescriptor:3.10"
+     xsi:schemaLocation="urn:proactive:jobdescriptor:3.10 http://www.activeeon.com/public_content/schemas/proactive/jobdescriptor/3.10/schedulerjob.xsd"
+    name="Cassandra" projectName="Cloud Automation - Deployment"
+    priority="normal"
+    onTaskError="continueJobExecution"
+     maxNumberOfExecution="2">
+  <variables>
+    <variable name="INSTANCE_NAME" value="cassandra-server" />
```
MaelAudren: This could be replace by a constant+an id. I'm not sure it brings value to let the user choose the name.

amouhoub: @MaelAudren Yes, definitely,,, if he ever wants to reuse an old docker container the he used in the past, or if for some reason the PCA job fails and the user wants to reuse the same docker container in order to not loose his config/data
@Aminelouati

PR #222, opened 2018-07-19
Fix cassandra database interaction fork env

amouhoub:
In this PR, we fix a fork env issue in `database-services/Cassandra_Database_Interaction` workflow


DatabaseServices/resources/catalog/Cassandra_Database_Interaction.xml
```diff
@@ -192,6 +197,11 @@ println("END " + variables.get("PA_TASK_NAME"))
         </script>
       </scriptExecutable>
       <controlFlow block="none"></controlFlow>
+      <metadata>
```
chloegugli: There is only positionLeft, I don't know why. So the studio will automatically generate a new position. I don't think it matters, no need to change it.

PR #225, opened 2018-07-23
PCA: Display docker container logs

amouhoub:
In this PR: we add a feature that allows to display logs of docker containers
We also fix the `Cassandra_Database_Interaction` workflow by adding a sleep(10000) to allow waiting for cassandra cluster to properly be up and listening for queries.

### DIFF
--- a/DatabaseServices/resources/catalog/Cassandra_Database_Interaction.xml
+++ b/DatabaseServices/resources/catalog/Cassandra_Database_Interaction.xml
@@ -661,11 +661,9 @@ preJavaHomeCmd = dockerRunCommand + dockerParameters + proActiveHomeVolume + wor
         <script>
           <code language="cpython">
             <![CDATA[
-import re
 import sys
+import time
 from cassandra.cluster import Cluster
-from cassandra.auth import PlainTextAuthProvider
-from cassandra.query import dict_factory
 
 HOSTNAME = variables.get("CASSANDRA_HOSTNAME")
 PORT = variables.get("CASSANDRA_PORT")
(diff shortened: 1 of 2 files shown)


DatabaseServices/resources/catalog/Cassandra_Database_Interaction.xml
```diff
@@ -680,6 +678,9 @@ if KEYSPACE is None:
     print("CASSANDRA_KEYSPACE not defined by the user.")
     sys.exit(1)
 
+# Wait for cassandra cluster to be up
+time.sleep(10)
```
nebilbenmabrouk: It is better to declare the value 10 as a variable, so it can be customized easily. This is useful when the host do not has Cassandra docker images yet and it takes longer to setup the the cluster.

Docker/resources/catalog/Docker.xml
```diff
@@ -96,6 +105,12 @@ while(currentStatus=="RUNNING") {
         serviceInstanceData = serviceInstanceRestApi.updateServiceInstanceUsingPUT(instanceId, serviceInstanceData)
         break
```
nebilbenmabrouk: Should add a sleep command so the  serviceInstance status is not fetched continuously. The wf will comsume less cpu if a sleep command is added. And it would be better that the sleep duration is declared as a variable.

DatabaseServices/resources/catalog/Cassandra_Database_Interaction.xml
amouhoub: Yes, Good idea

Docker/resources/catalog/Docker.xml
amouhoub: In fact there is a sleep of 10s, first line inside the loop. (Line 91)

PR #228, opened 2018-08-12
Fix PCA collision and add logs feature

amouhoub:
In this PR, we propose a fix for the collision bug between PCA services that creates a deadlock if all nodes are busy running PCA services. This fix unleashes the load on the available nodes by running PCA updates as cron tasks every 1 minute

We also add a feature that fetches displays docker container logs that the PCA rely on

### DIFF
--- a/Cassandra/resources/catalog/Resume_Cassandra.xml
+++ b/Cassandra/resources/catalog/Resume_Cassandra.xml
@@ -13,13 +13,105 @@
   <genericInformation>
     <info name="bucketName" value="cloud-automation"/>
     <info name="workflow.icon" value="/automation-dashboard/styles/patterns/img/wf-icons/cassandra.png"/>
-    <info name="pca.states" value="(STOPPED,RUNNING)(RUNNING,RUNNING)"/>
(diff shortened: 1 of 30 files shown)


Cassandra/resources/catalog/Resume_Cassandra.xml
```diff
@@ -13,13 +13,97 @@
...
+      </description>
+      <genericInformation>
+        <info name="task.icon" value="/automation-dashboard/styles/patterns/img/wf-icons/cassandra.png"/>
+        <info name="Documentation" value="https://doc.activeeon.com/latest/PCA/PCAUserGuide.html"/>
+      </genericInformation>
+      <depends>
+        <task ref="Resume_Cassandra"/>
+      </depends>
+      <inputFiles>
+        <files  includes="cloud-automation-service-client-8.2.0-SNAPSHOT.jar" accessMode="transferFromGlobalSpace"/>
```
paraita: is it normal that the version is hardcoded ?

amouhoub: @paraita Yes,,, unfortunately unless we rename it elsewhere before importing it in the global space but that's a very good point. I'll see what can be done on the other side...

PR #251, opened 2018-09-05
Add collection of selection scripts

fviale:
- covers a wide range of selection script usages
  - additionally change kind associated with kill_other_replicated_tasks
  - additionally, use variables in start-at example


SelectionScripts/resources/catalog/is_windows.groovy
```diff
@@ -0,0 +1,19 @@
+/**
+ * Script which verifies that the current node runs in a windows machine
+ */
+
+import org.ow2.proactive.utils.OperatingSystem;
+import org.ow2.proactive.utils.OperatingSystemFamily;
+
+String osName = System.getProperty("os.name");
+println "Node operating system : " + osName;
+OperatingSystem operatingSystem = OperatingSystem.resolveOrError(osName);
```
nebilbenmabrouk: The class 'org.ow2.proactive.utils.OperatingSystem' does not consider Windows 10. It should  be updated accordingly.

build.gradle
```diff
@@ -84,6 +84,7 @@ task zip (type: Zip){
     include 'RemoteVisualization/**'
     include 'ReserveNodes/**'
     include 'Scripts/**'
+    include 'SelectionScripts/**'
```
nebilbenmabrouk: Global comments:
1) We have two classes named 'OperatingSystem'. 
- org.ow2.proactive.utils.OperatingSystem
- org.ow2.proactive.resourcemanager.utils.OperatingSystem
Is it possible to use only one and remove the other?
2) We do not have a selection script that uses (recent feature of) the node URL or hostname or ip address. Is it possible to add such scripts to the catalog?

SelectionScripts/resources/catalog/is_windows.groovy
fviale: I agree, in that case it is treated as "Windows Generic", but we should update this class with Windows 10

build.gradle
fviale: 1. I can have a look if we can easily remove the unnecessary class (we have to take care of gradle project scope).

2) the selection scripts I wrote try to use these features when appropriate, for example the check_node_name[_regexp], use the nodename binding. The check_host_name[_regexp] use the nodehost binding.

I did not create a selection script to select a node by url, as nodes urls are very volatile, will typically change upon restart, etc. Creating such a selection script is simple thanks to the nodeurl binding, but incorporate such a script in the default library makes not much sense.

PR #280, opened 2018-10-15
openshiftExamples

tobwiens:


tobwiens:
I would prefer to not coy the workflows but just reference them for better maintainability. If any reviewer know how to do it, that would be great.

OpenShift/METADATA.json
```diff
@@ -0,0 +1,48 @@
...
+  "metadata": {
+    "slug": "OpenShift",
+	"name": "OpenShift Examples",
+	"short_description": "Examples which are OpenShift compatible and relevant",
+	"author": "ActiveEon's Team",
+	"tags": [
+	  "OpenShift"
+	],
+	"version": "1.0"
+  },
```
Aminelouati: I think that you forgot to precise the test section
```"test": {
    "scenarios": "resources/test/scenarios.json"
  },```

tobwiens: Yes you are right. Added tests :+1:

PR #283, opened 2018-10-17
openshiftExamples

tobwiens:


OpenShift/resources/test/scenarios.json
```diff
@@ -1,4 +1,20 @@
...
+    "workflow":"CriticalSection",
+    "variables": {}
+  },
+  {
+    "workflow":"CriticalSectionWithSelectionScripts",
+    "variables": {}
+  },
   {
     "workflow":"distributed_computing_PI",
     "variables": {}
```
Aminelouati: Not sure but I think that you can put more scenarios here like:  print_file_name and data_management

tobwiens: I wasn't sure either. I didn't see those scenarios in the other package. So I figure, that print-file name -> actually needs a file name to succeed. 

data_management <- I will add it

tobwiens:
@Aminelouati I added data_management to the scenarios

tobwiens:
@luispinedamx You are right a naming convention would be great. The different naming comes from me copying the workflows from other packages.

tobwiens:
@luispinedamx I will leave the naming as it is, because I like to keep the "history" of the workflows. I hope that if one changes workflow x -> that one then sees that this workflow (with the exact same name) exists in other packages -> so that person will also change the other versions with it.

PR #286, opened 2018-10-19
GPN : adding workflow ocr in basic examples bucket

gparanthoen:
adding workflow ocr in bucket examples, 

this workflow will work at the moment on try only due to a selection script that target try nodes OR OCRnodesource.

### DIFF
--- a/Demo/METADATA.json
+++ b/Demo/METADATA.json
@@ -1,0 +1,24 @@
+{
+	"metadata": {
+		"slug": "demo",
+		"name": "Demo",
+		"short_description": "Demo",
+		"author": "ActiveEon's Team",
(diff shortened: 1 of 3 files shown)


Demo/METADATA.json
```diff
@@ -0,0 +1,24 @@
+{
+	"metadata": {
+		"slug": "demo",
```
vinseon: "demo" is too generic, better to call it OCR or Tesseract instead

build.gradle
```diff
@@ -98,6 +98,7 @@ task zip (type: Zip){
     include 'WebNotification/**'
     include 'Zookeeper/**'
     include 'OpenShift/**'
+    include 'Demo/**'
```
vinseon: Same comment than before (OCR or Tesseract would be a better name)

PR #318, opened 2018-12-11
Add test scenarios to all eligible packages

amouhoub:
In this PR, we add test scenarios to all eligible packages.
We also add some workflow tests in existing scenarios

### DIFF
--- a/GetStarted/resources/test/scenarios.json
+++ b/GetStarted/resources/test/scenarios.json
@@ -10,5 +10,21 @@
   {
     "workflow":"Variables_Propagation",
     "variables":{}
+  },
+  {
+    "workflow":"Multi_Node_Task",
+    "variables":{}
+  },
+  {
+    "workflow":"Native_Task",
+    "variables":{}
+  },
+  {
+    "workflow":"Pre_Post_Clean_Scripts",
+    "variables":{}
+  },
+  {
+    "workflow":"Selection_Scripts",
+    "variables":{}
   }
 ]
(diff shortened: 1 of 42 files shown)


GetStarted/resources/test/scenarios.json
```diff
@@ -10,5 +10,21 @@
...
+  },
+  {
+    "workflow":"Multi_Node_Task",
+    "variables":{}
+  },
+  {
+    "workflow":"Native_Task",
+    "variables":{}
+  },
+  {
```
luispinedamx: What about `Native_Task_Linux` and `Native_Task_Windows`? they are not eligible? Honest question, I really don't know.

MachineLearningH2O/resources/test/scenarios.json
```diff
@@ -0,0 +1,6 @@
+[
+  {
+    "workflow":"Test_H2O_Service",
+    "variables": {}
+  }
```
luispinedamx: Same here, what about `Start_H2O_Service`, `Stop_H2O_Service` and `Finish_H2O_Service`?
In all honesty this is more a mechanic review rather than a semantic one :)

amouhoub: in fact, `Test_H2O_Service` uses the others which are task-workflows and which cannot work properly alone but all together as demonstrated in `Test_H2O_Service`

GetStarted/resources/test/scenarios.json
amouhoub: `Native_Task_Linux` is just above at line 7
`Native_Task_Windows` will not work because system-tests-only is a linux machine, therefore the workflow will hang forever because of its selection script or will fail.

PR #347, opened 2019-01-14
Fix visdom container collision issue

amouhoub:


Aminelouati:
Just for info, I found 24 matches of `visdom-server` in the project and you changed only 6 files.

amouhoub:
> Just for info, I found 24 matches of `visdom-server` in the project and you changed only 6 files.

@Aminelouati I'm aware of that but I only changed those that failed. The others use a different name (visdom-server-1). They could fail as well but I didn't change them.
I think that applies to all other workflows and services (database, Kafka, storm, ELK, etc,,,)

PR #367, opened 2019-02-13
Updated description field

carolinepacheco:
small modifications.


carolinepacheco:
updated description field.

PR #378, opened 2019-02-21
Fix some bugs

ImenBizid:
List of bugs
1. Machine Leaning Bucket
     1.1  Fill_NanNs task: has python Language instead cpyton Language.
2. Machine Learning Workflows Bucket
     2.1 Diabetics_Detection_Using_K_Means workflow: the Download_Model task doesn't  download model correctly (zero bytes).
     2.2 California_Housing_Prediction_Using_TPOT_Regressor: it takes a long time to run.


MLBasics/resources/catalog/California_Housing_Prediction_Using_TPOT_Regressor.xml
```diff
@@ -74,11 +76,13 @@ import pandas as pd
 import numpy as np
 
 from sklearn.datasets import fetch_california_housing
+from sklearn.model_selection import train_test_split
 california_housing = fetch_california_housing()
 dataframe_load = pd.DataFrame(california_housing.data)
 dataframe_load.columns = california_housing.feature_names 
 data_label = california_housing.target
 dataframe = dataframe_load.assign(LABEL=data_label)
+dataframe, neglected = train_test_split(dataframe, test_size=0.98)
```
carolinepacheco: Why a train_test_split in the Load_Californa_Housing_Dataset task?  not clear...

MLBasics/resources/catalog/Diabetics_Detection_Using_K_Means.xml
```diff
@@ -1,27 +1,29 @@
-<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
carolinepacheco: Whats the difference from this workflow and the previous one?

MLBasics/resources/catalog/California_Housing_Prediction_Using_TPOT_Regressor.xml
ImenBizid: it is used to reduce the dataset size as the full dataset is huge.

MLBasics/resources/catalog/Diabetics_Detection_Using_K_Means.xml
ImenBizid: just corrected the the link between train model and download model.

PR #398, opened 2019-03-22
Add new cloud watch rules

ShatalovYaroslav:
Please ignore the commits amount, it will be squashed during the merge


PCWRules/resources/catalog/File_System_Used_Disk_Space_Percent_Rule.xml
```diff
@@ -42,12 +42,11 @@ rule #autogenerated_rule_id
...
   exists  kpi : NodeMetrics( url == "#nodeUrlVar" )
   $result : NodeMetrics($metricTotal : metrics["#totalSizeFSMetric"])
- NodeMetrics($metricUsed : metrics["#usedSizeFSMetric"] / $metricTotal * 100 > #percentageUsedLimit)
+  NodeMetrics($metricUsed : metrics["#usedSizeFSMetric"] / $metricTotal * 100 > #percentageUsedLimit)
 then
-    System.out.println("In then condition of rule for file system limit, amount of used: " + $metricUsed);
     Map<String, String> parameters = new HashMap();
 	parameters.put("credentialsPath","#pathToCredentialFileVar");
-    executorWorkflowFromCatalog.submitToScheduler(parameters, "#autogenerated_rule_id", "{'bucket_name':'#workflowNameAsActionVar','workflow_name':'#workflowNameAsActionVar','variables':{}}");
+    executorWorkflowFromCatalog.submitToScheduler(parameters, "#autogenerated_rule_id", "{'bucket_name':'#bucketNameOfWorkflowAsActionVar','workflow_name':'#workflowNameAsActionVar','variables':{}}");
```
aminelouati: It is kind of a dependency between a rule and a workflow. May be you can add a variable having the model PA:Catalog_Object and specify the dependency (bucket_name/workflow_name) in the value.  We can then improve PCWRuleParser to include such information.

ShatalovYaroslav: Very good point. I need to investigate this.

amouhoub: What a coincidence. @nxnarbais has yielded such a requirement yesterday to know if jobs were ran manually, automatically from other workflows or automatically from cloud-watch rules. That would be very interesting to have and even more to display it in the scheduler portal !!!

PCWRules/resources/catalog/CPU_Monitoring_Average_Time_Window_Rule.xml
```diff
@@ -0,0 +1,52 @@
...
+                <parameter name="jmxCredentialsFile" value="#pathToCredentialFileVar" />
+                <parameter name="nodeUrl" value="#nodeUrlVar" />
+                <parameter name="pollingPeriodInSeconds" value="#pollingPeriodInSecondsVar" />
+                <parameter name="calmPeriodInSeconds" value="#calmPeriodInSecondsVar" />
+            </parameters>
+        </pollingClass>
+        <kpis> <kpi>#usedCPUMetric</kpi> </kpis>
+        <droolContent>
+            <![CDATA[
+package org.ow2.proactive.cloud_watch.rules
```
amouhoub: Maybe reformat the code bellow to use 4-space indentations (40-43) and remove preceding spaces on lines 31 and 32

PCWRules/resources/catalog/CPU_Usage_Percent_Monitoring_Rule.xml
```diff
@@ -0,0 +1,54 @@
...
+                <parameter name="calmPeriodInSeconds" value="#calmPeriodInSecondsVar" />
+            </parameters>
+        </pollingClass>
+        <kpis>
+            <kpi>#totalCPUMetric</kpi>
+            <kpi>#usedCPUMetric</kpi>
+        </kpis>
+        <droolContent>
+            <![CDATA[
+ package org.ow2.proactive.cloud_watch.rules
```
amouhoub: Same as above

PCWRules/resources/catalog/Memory_Free_Monitoring_Rule.xml
```diff
@@ -0,0 +1,49 @@
...
+                <parameter name="jmxCredentialsFile" value="#pathToCredentialFileVar" />
+                <parameter name="nodeUrl" value="#nodeUrlVar" />
+                <parameter name="pollingPeriodInSeconds" value="#pollingPeriodInSecondsVar" />
+                <parameter name="calmPeriodInSeconds" value="#calmPeriodInSecondsVar" />
+            </parameters>
+        </pollingClass>
+        <kpis> <kpi>#freeMemoryMetric</kpi> </kpis>
+        <droolContent>
+            <![CDATA[
+package org.ow2.proactive.cloud_watch.rules
```
amouhoub: Same as above

PCWRules/resources/catalog/Node_Is_Down_Cloud_Automation_Action_Rule.xml
```diff
@@ -0,0 +1,47 @@
...
+                <parameter name="jmxCredentialsFile" value="#pathToCredentialFileVar" />
+                <parameter name="nodeUrl" value="#nodeUrlVar" />
+                <parameter name="pollingPeriodInSeconds" value="#pollingPeriodInSecondsVar" />
+                <parameter name="calmPeriodInSeconds" value="#calmPeriodInSecondsVar" />
+            </parameters>
+        </pollingClass>
+        <kpis> <kpi>upAndRunning</kpi> </kpis>
+        <droolContent>
+            <![CDATA[
+package org.ow2.proactive.cloud_watch.rules
```
amouhoub: Same as above

PCWRules/resources/catalog/File_System_Used_Disk_Space_Percent_Rule.xml
ShatalovYaroslav: @aminelouati @amouhoub Thank you for your suggestion. I've added this point to PCW epic. The only problem and difference that I see here: that this association between rule template and workflow is just example for user. This rule template is stored in catalog. But when a user will actually activate rule the value of (bucket_name/workflow_name) normally will be changed without saving in catalog. So the real assocaition between activated rule and wf will be stored only in PCW service.

In any case I will investigate about this idea.

ShatalovYaroslav:
retest this please

PCWRules/resources/catalog/CPU_Usage_Percent_Monitoring_Rule.xml
ShatalovYaroslav: the code reformat

PR #399, opened 2019-04-02
Openstack nodes sources added

nebilbenmabrouk:


NodeSources/METADATA.json
```diff
@@ -83,13 +83,31 @@
...
 			{
-				"name" : "OpenStack",
+				"name" : "OVH",
 				"metadata" : {
 					"kind": "NodeSource",
 					"commitMessage": "First commit",
 					"contentType": "application/json"
 				},
-				"file" : "resources/catalog/OpenStack.json"
+				"file" : "resources/catalog/OVH_static_policy.json"
```
mcompastie: This file is not present in the pull request. Is that normal ?

nebilbenmabrouk: No it is present. It is the second file added in the PR

PR #434, opened 2019-06-05
add GoogleComputeEngine nodesource example.

tao-xinxiu:


NodeSources/resources/catalog/GoogleCloudCompute.json
```diff
@@ -1,35 +0,0 @@
-{
-  "nodeSourceName": "GoogleCloudCompute",
```
MaelAudren: I think that this nodesource description is currently used in try as a mock. Did you ensure that you will not break something by removing this nodesource ?

tao-xinxiu: After verified with Fabien and checking deployment scripts, the fake nodesources on try are deployed by `deploy_fake_nodesource` through client CLI which directly specifies the infrastructure and policy without using this template. So, this file will not impact the startup behavior.

PR #440, opened 2019-06-11
data-visualization tasks fork env as a reference, fixed code format and file location

andrewssobral:


andrewssobral:
Hello @fviale , in fact the `MachineLearningScripts/resources/catalog/Visdom_Visualize_Results.py` is the correct file, but you are right, the correct is to move this file to a new folder named DataVisualizationScripts, I will do it in the next PR.

andrewssobral:
@fviale for the fork env, what do you prefer, move fork_env.py to Scripts or create a new directory named ForkEnvironmentScripts ?

fviale:
@andrewssobral but I understood this pull request was targeting at changing inline fork environment script to reference one. But I see many changes in a visdom script which is not a fork environment one.

Additionally, the many code formatting changes makes the visibility difficult.

Regarding script placement, I was rather pointing to the script : ${PA_CATALOG_REST_URL}/buckets/scripts/resources/fork_env/raw
which is stored in the proactive examples project inside the SelectionScript module (which is a mistake).

I agree that the last bit could be a separate commit or PR. But the PR itself should be cleaned to remove all the many format changes.

fviale:
I think it's ok to move it to Scripts

andrewssobral:
@fviale  ok.
Yes, I am doing some code formatting by changing from 2 spaces to 4 spaces, I need to do this to be compatible with the PEP8 standard `https://www.python.org/dev/peps/pep-0008/`, we have several python scripts with 2 spaces, this is an old mistake that should be fixed. Moreover, the script editor from studio is considering indentation with 4 spaces by default.

fviale:
All right, I understand, and remove my warning accordingly.

andrewssobral:
All workflows modified in this PR were tested with a fresh nightly-release by building the proactive-examples from scratch and replacing the proactive-examples.zip file in the samples folder. The visualization worked fine, as well the workflow executions.
Screen Shot 2019-06-11 at 23 23 48

PR #442, opened 2019-06-14
Fix some issues in NLP tasks and workflows

ImenBizid:


andrewssobral:
I think the best is to commit (or create a PR) the workflows using the same approach every time, for ex:
1. by exporting the workflow as XML in the studio.
or
2. by pushing the workflow from the studio to the catalog, then going to automation-dashboard and downloading the respective workflow.
Unfortunately the produced XML is a little bit different between approach 1 and 2, so this is the cause every time we have a lot of modified lines that does not comes from the user modifications itself. In short, you change only 1 variable and the commit shows 100s lines of modifications, this make the review so boring every time.

PR #450, opened 2019-06-27
Improve workflow templates for Connection Pools and correct Email and Web Notification task names

aminelouati:


DataConnectors/resources/catalog/SQL_Pooled_Connection_Update.xml
```diff
@@ -4,10 +4,23 @@
   xmlns="urn:proactive:jobdescriptor:3.11" xsi:schemaLocation="urn:proactive:jobdescriptor:3.11 http://www.activeeon.com/public_content/schemas/proactive/jobdescriptor/3.11/schedulerjob.xsd"  name="SQL_Pooled_Connection_Update" projectName="2. SQL" priority="normal" onTaskError="continueJobExecution"  maxNumberOfExecution="2" >
   <variables>
     <variable name="RDBMS_NAME" value="HSQLDB" model="PA:LIST(PostgreSQL, MySQL, Greenplum, SQL Server, Oracle, HSQLDB)"/>
+    <variable name="HOST" value="localhost" />
+    <variable name="PORT" value="" />
+    <variable name="USERNAME" value="scheduler" />
```
amouhoub: maybe change the default value of USERNAME with empty values as they will definitely be changed by the user (in contrast with localhost for HOST which will most probably be used as is)

DataConnectors/resources/catalog/SQL_Pooled_Connection_Update.xml
```diff
@@ -4,10 +4,23 @@
...
+    <variable name="DATABASE" value="scheduler" />
```
amouhoub: Same as above

DataConnectors/resources/catalog/SQL_Pooled_Connection_Update.xml
aminelouati: you are right. However, for SQL_Pooled_Connection_Query.xml do you agree to keep the default values to make the template executable ?

DataConnectors/resources/catalog/SQL_Pooled_Connection_Query.xml
```diff
@@ -116,132 +124,140 @@ if (!password) {
...
+    dBConnectionPoolsHolder = DBConnectionPoolsHolder.getInstance();
+    rs = dBConnectionPoolsHolder.executeQuery(dbConnectionDetails, sqlStatements)
+} catch (Exception e) {
+    e.printStackTrace();
+    throw e;
+}
+interceptor.stop()
 
+OUTPUT_TYPE = variables.get("OUTPUT_TYPE")
+OUTPUT_FILE = variables.get("OUTPUT_FILE")
```
fviale: this variable is not used.

DataConnectors/resources/catalog/SQL_Pooled_Connection_Query.xml
```diff
@@ -116,132 +124,140 @@ if (!password) {
...
 sqlStatements = variables.get("SQL_STATEMENTS")   
 if (!sqlStatements){
     throw new IllegalArgumentException("ERROR: SQL_STATEMENTS variable is not provided by the user. Empty value is not allowed.")
 }
+ResultSet rs = null
+try {
+    dBConnectionPoolsHolder = DBConnectionPoolsHolder.getInstance();
+    rs = dBConnectionPoolsHolder.executeQuery(dbConnectionDetails, sqlStatements)
+} catch (Exception e) {
+    e.printStackTrace();
```
fviale: I am afraid this printStackTrace works only in fork mode.

in non-fork mode:
println "something" works
System.out.println("something") does not work, and printStackTrace() I believe is the same.

Could you check ?

aminelouati: yes in non fork mode printStackTrace() does not work

DataConnectors/resources/catalog/SQL_Pooled_Connection_Update.xml
amouhoub: @aminelouati yeah sure, I was just talking about `USERNAME` and `DATABASE`

PR #452, opened 2019-06-28
enhance connection pools workflow template

aminelouati:


DataConnectors/resources/catalog/SQL_Pooled_Connection_Query.xml
```diff
@@ -184,6 +182,26 @@ if (outputType == "HTML"){
     resultMetadata.put("content.type", "text/csv")
 }
 
+storeResultVariable = variables.get("STORE_RESULT_VARIABLE")
+if(storeResultVariable){
+    try {
+        MapListHandler handler = new MapListHandler() {
+            public List<Map<String, Object>> handle(ResultSet resultSet) throws SQLException {
+                List results = new ArrayList();
```
mboussaa: ; to remove?

aminelouati: yep

DataConnectors/resources/catalog/SQL_Pooled_Connection_Query.xml
```diff
@@ -184,6 +182,26 @@ if (outputType == "HTML"){
     resultMetadata.put("content.type", "text/csv")
 }
 
+storeResultVariable = variables.get("STORE_RESULT_VARIABLE")
+if(storeResultVariable){
+    try {
+        MapListHandler handler = new MapListHandler() {
+            public List<Map<String, Object>> handle(ResultSet resultSet) throws SQLException {
```
fviale: I don't think it's necessary to implement the method. I believe it's already implemented in the MapListHandler class or parent class. To double check

PR #522, opened 2019-11-26
New Rest template tasks and bucket

fviale:
- templates to perform various kind of rest calls (Get, put, post, delete, patch)
 - with either body, form, query params, etc
 - restassured library is used


Rest/resources/catalog/Body_Delete.xml
```diff
@@ -0,0 +1,193 @@
...
+        <variable name="USER" value="" inherited="true" />
+        <variable name="PASSWORD" value="" inherited="true" model="PA:HIDDEN"/>
+        <variable name="PATH" value="/delete" inherited="false" />
+        <variable name="SSL_DISABLE_CHECK" value="true" inherited="false" model="PA:BOOLEAN"/>
+        <variable name="REQUEST_BODY" value="{ &quot;foo1&quot;:&quot;bar1&quot; }" inherited="false" model="PA:JSON"/>
+        <variable name="DEBUG" value="false" inherited="false" model="PA:BOOLEAN"/>
+        <variable name="RESPONSE_FORMAT" value="json" inherited="false" model="PA:LIST(string,json,xml,html,contentView)"/>
+        <variable name="RESPONSE_PATH" value="data.foo1" inherited="false" />
+        <variable name="HEADER_head1" value="val1" inherited="false" />
+        <variable name="QUERYPARAM_foo2" value="bar2" inherited="false" />
```
MaelAudren: It should be QUERY_PARAM_... instead of QUERY_PARAM....
Moreover I think that if we want to unified with the HEADER line. It would be better to have ```name="QUERY_PARAM_param1"  value="val2"```

Rest/resources/catalog/Body_Put.xml
```diff
@@ -0,0 +1,193 @@
...
+    <info name="workflow.icon" value="/automation-dashboard/styles/patterns/img/wf-icons/api-rest.png"/>
+  </genericInformation>
+  <taskFlow>
+    <task name="RestBodyPut" 
+    
+    
+    
+    
+    fork="true"
>
```
mboussaa: what is ^M?

Rest/resources/catalog/Body_Delete.xml
```diff
@@ -0,0 +1,193 @@
...
+
+if (!variables.get("USER").isEmpty() && variables.get("PASSWORD").isEmpty()) {
+    restCall = restCall.auth().preemptive().basic(variables.get("USER"), variables.get("PASSWORD"))
+}
+
+variables.entrySet().stream().filter({entry -> entry.getKey().startsWith("HEADER_")}).forEach({entry -> restCall = restCall.header(entry.getKey().replace("HEADER_",""), entry.getValue()) });
+
+variables.entrySet().stream().filter({entry -> entry.getKey().startsWith("QUERYPARAM_")}).forEach({entry -> restCall = restCall.queryParam(entry.getKey().replace("QUERYPARAM_",""), entry.getValue()) });
+
+if (variables.get("REQUEST_BODY") == null || variables.get("REQUEST_BODY").isEmpty()) {
```
MaelAudren: It is weird to have the body required for a delete request. Why did you do that ?

Rest/resources/catalog/Body_Delete.xml
```diff
@@ -0,0 +1,193 @@
...
+    <info name="workflow.icon" value="/automation-dashboard/styles/patterns/img/wf-icons/api-rest.png"/>
+  </genericInformation>
+  <taskFlow>
+    <task name="RestBodyDelete" 
+    
+    
+    
+    
+    fork="true"
>
```
aminelouati: Can you please remove Windows line-ending character ^M ?

Rest/resources/catalog/Body_Patch.xml
```diff
@@ -0,0 +1,193 @@
...
+    <info name="workflow.icon" value="/automation-dashboard/styles/patterns/img/wf-icons/api-rest.png"/>
+  </genericInformation>
+  <taskFlow>
+    <task name="RestBodyPatch" 
+    
+    
+    
+    
+    fork="true"
>
```
aminelouati: same here

Rest/resources/catalog/Body_Post.xml
```diff
@@ -0,0 +1,193 @@
...
+    <info name="workflow.icon" value="/automation-dashboard/styles/patterns/img/wf-icons/api-rest.png"/>
+  </genericInformation>
+  <taskFlow>
+    <task name="RestBodyPost" 
+    
+    
+    
+    
+    fork="true"
>
```
aminelouati: same here

Rest/resources/catalog/Body_Delete.xml
fviale: because it is allowed by the specification and it is totally valid. Body are simply parameters given to the delete request

Rest/resources/catalog/Body_Delete.xml
fviale: of for changing the naming

Rest/resources/catalog/Body_Delete.xml
fviale: sure

fviale:
@aminelouati 
thanks to point this out. The ordered_bucket_list seems rather outdated, though

Rest/resources/catalog/Body_Delete.xml
MaelAudren: Sure, what I mean is that I think that REQUEST_BODY should not be mandatory but optional.

fviale: Ahh, damn copy paste, you are right, and thanks for spotting this out!
Do you agree that the body should not be empty for POST/PUT/PATCH requests ?

PR #548, opened 2019-12-18
Improve Secure PCA use for Tensorboard and Visdom

ImenBizid:
In this PR, we add the following enhancements for Visdom and Tensorboard PCA services:
- add a boolean variable `PROXYFIED` in order to be able to choose between PCA secure or not
- update the client PCA tasks
- update each service actions to be able to detect automatically the service instance_name
- delete the Visdom_Secure Service
- adapt the endpoint map


Tensorboard/resources/catalog/Tensorboard.xml
```diff
@@ -5,6 +5,7 @@
   <variables>
     <variable name="INSTANCE_NAME" value="tensorboard-server" />
     <variable name="MOUNT_LOG_PATH" value="/shared/$INSTANCE_NAME" />
+    <variable name="PROXYFIED" value="true" model="PA:Boolean"/>
```
andrewssobral: @ImenBizid ... in the `Visdom_Service_Start.xml` you put `<variable name=“PROXYFIED” value=“True” model=“PA:Boolean”/>` (value="True") but here you put `<variable name="PROXYFIED" value="true" model="PA:Boolean"/>` (value="true").... make sure when you do `if (proxyfied=="true"){` it will not work when the value is "True" instead of "true".

Visdom/resources/catalog/Visdom.xml
```diff
@@ -121,16 +145,49 @@ def apiClient = new ApiClient()
...
+// Update service instance data : (Container, Endpoint, Deployment)
+def Container container = new Container();
+container.setId(containerID);
+container.setName(instanceName);
+def Endpoint endpoint = new Endpoint();
+endpoint.setId(endpointID);
+endpoint.setUrl(containerUrl);
+
+// Set the endpoint according to the Proxy settings
+if (proxyfied=="true"){
```
andrewssobral: same here...

Visdom/resources/catalog/Visdom.xml
```diff
@@ -175,31 +232,31 @@ It will run every minute. ]]>
         <task ref="Start_Visdom"/>
       </depends>
       <inputFiles>
-        <files accessMode="transferFromGlobalSpace" includes="cloud-automation-service-client-8.2.0-SNAPSHOT.jar"/>
+        <files  includes="cloud-automation-service-client-8.2.0-SNAPSHOT.jar" accessMode="transferFromGlobalSpace"/>
```
andrewssobral: not sure....i think you can remove this....
this is an old workaround

Visdom/resources/catalog/Visdom.xml
```diff
@@ -175,31 +232,31 @@ It will run every minute. ]]>
...
       </inputFiles>
-      <forkEnvironment>
+      <forkEnvironment >
         <additionalClasspath>
           <pathElement path="cloud-automation-service-client-8.2.0-SNAPSHOT.jar"/>
```
andrewssobral: same here...

Visdom/resources/catalog/Visdom.xml
```diff
@@ -121,16 +145,49 @@ def apiClient = new ApiClient()
...
+    serviceURL = proxyfiedURL
+}else{
+    endpoint.setProxyfied(false);
+	endpoint.setProxyfiedUrl(null); 
+    serviceURL = containerUrl
+}
+
 //Update endpoints
 def endpointsMap = [:]
 endpointsMap.put(instanceName, endpoint)
```
nebilbenmabrouk: Should be `endpointsMap.put(instanceName,serviceURL`

ImenBizid: Yes you are right

Visdom/resources/catalog/Visdom.xml
```diff
@@ -121,16 +145,49 @@ def apiClient = new ApiClient()
...
+container.setName(instanceName);
+def Endpoint endpoint = new Endpoint();
+endpoint.setId(endpointID);
+endpoint.setUrl(containerUrl);
+
+// Set the endpoint according to the Proxy settings
+if (proxyfied=="true"){
+    endpoint.setProxyfied(true);
+	endpoint.setProxyfiedUrl(proxyfiedURL);
+    serviceURL = proxyfiedURL
```
nebilbenmabrouk: Don't we need to declare the variable `serviceURL` above?

Visdom/resources/catalog/Visdom.xml
```diff
@@ -121,16 +145,49 @@ def apiClient = new ApiClient()
...
+//node.setHost(variables.get("PA_NODE_HOST"))
+def Deployment deployment = new Deployment();
+//deployment.setNode(node);
+deployment.setContainer(container);
+deployment.setEndpoint(endpoint);
+
+// Update service instance data : (legacy endpoint, Deployment, Groups)
 def serviceInstanceData = serviceInstanceRestApi.getServiceInstanceUsingGET(instanceId)
 serviceInstanceData.setInstanceStatus("RUNNING")
+serviceInstanceData = serviceInstanceData.putInstanceEndpointsItem(instanceName, serviceURL);
```
ShatalovYaroslav: Please don't use any more "putInstanceEndpointsItem"
it is deprecated method.
Just remove this line, as the put is already done in serviceInstanceRestApi.createNewInstanceEndpointsUsingPUT

Visdom/resources/catalog/Visdom.xml
ImenBizid: I think, no need it works correctly like that

Tensorboard/resources/catalog/Tensorboard.xml
```diff
@@ -157,8 +162,23 @@ container.setName(instanceName);
...
 endpoint.setId(endpointID);
 endpoint.setUrl(containerUrl);
-endpoint.setProxyfied(true);
-endpoint.setProxyfiedUrl(proxifiedURL);
+
+// Set the endpoint according to the Proxy settings
+if (proxyfied=="true"){
+    endpoint.setProxyfied(true);
+	endpoint.setProxyfiedUrl(proxyfiedURL);
+    serviceURL = proxyfiedURL
```
ShatalovYaroslav: Could you move here the code ?
`
//Update endpoints
def endpointsMap = [:]
endpointsMap.put(instanceName, proxyfiedURL)
serviceInstanceRestApi.createNewInstanceEndpointsUsingPUT(instanceId, endpointsMap)
`
In this case we will not have 2 times in portal non proxified Url

Tensorboard/resources/catalog/Tensorboard.xml
ImenBizid: Right, I will fix it

Visdom/resources/catalog/Visdom.xml
ImenBizid: Yes, it is the old version

Visdom/resources/catalog/Visdom.xml
```diff
@@ -121,16 +145,49 @@ def apiClient = new ApiClient()
...
+container.setId(containerID);
+container.setName(instanceName);
+def Endpoint endpoint = new Endpoint();
+endpoint.setId(endpointID);
+endpoint.setUrl(containerUrl);
+
+// Set the endpoint according to the Proxy settings
+if (proxyfied=="true"){
+    endpoint.setProxyfied(true);
+    endpoint.setProxyfiedUrl(proxyfiedURL);
```
ShatalovYaroslav: same comment as before
Could you move here the code ?
`//Update endpoints def endpointsMap = [:] endpointsMap.put(instanceName, proxyfiedURL) serviceInstanceRestApi.createNewInstanceEndpointsUsingPUT(instanceId, endpointsMap)`

Visdom/resources/catalog/Visdom.xml
nebilbenmabrouk: OK

Visdom/resources/catalog/Visdom.xml
ImenBizid: We will keep only one endpoint in the map for this version. No need to have both URLs

Visdom/resources/catalog/Visdom.xml
ImenBizid: We will discuss about that tomorrow with Fabien as we need to keep this jar in order to be able to execute the services on distant nodes.

fviale: Yes, I always have near heart attacks when I see this line !

PR #557, opened 2019-12-26
fix version resolving erro

aminelouati:


build.gradle
```diff
@@ -137,14 +137,14 @@ configurations {
 
 dependencies {
     awsJar 'com.amazonaws:aws-java-sdk-s3'
```
nebilbenmabrouk: @aminelouati please add hardcoded version here also, to be sure that is will work. Thanks

PR #583, opened 2020-02-12
remove '$' from the name of variables in the description tag

ShatalovYaroslav:
According to Scheduler workflow parser: all the variables represented in format $var are replaced by the variable values inside the whole workflow content (in the description and other parts of the workflow). This behavior gives issue in showing variable names in description of the workflow in the portals.
 So the solution is to remove '$' from the name of variables in the description tag

### DIFF
--- a/Docker/resources/catalog/Docker.xml
+++ b/Docker/resources/catalog/Docker.xml
@@ -19,11 +19,11 @@ The underlying docker command is the following:
 docker run -- name $DOCKER_CONTAINER_NAME -p $DOCKER_PORT $DOCKER_OPTIONS -d $DOCKER_IMAGE $DOCKER_COMMAND
 
 The service can be started using the following variables:
-$DOCKER_IMAGE: Docker image name. It can include a tag as well.
-$DOCKER_PORT: The main image port. Please note that it will be forwarded to a random port that will be returned by this service.
-$DOCKER_CONTAINER: If you desire to stop this container and restart it later.
-$DOCKER_OPTIONS (OPTIONAL): options like environment variables, etc.
-$DOCKER_COMMAND (OPTIONAL) ]]>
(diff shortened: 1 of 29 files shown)


Docker/resources/catalog/Docker.xml
```diff
@@ -293,11 +293,11 @@ The underlying docker command is the following:
...
+DOCKER_IMAGE: Docker image name. It can include a tag as well.
+DOCKER_PORT: The main image port. Please note that it will be forwarded to a random port that will be returned by this service.
+DOCKER_CONTAINER: If you desire to stop this container and restart it later.
+DOCKER_OPTIONS (OPTIONAL): options like environment variables, etc.
+DOCKER_COMMAND (OPTIONAL)"><img src="/automation-dashboard/styles/patterns/img/wf-icons/docker.png" width="20px">&nbsp;<span class="name">Start_Docker</span></a></div><div class="task ui-draggable _jsPlumb_endpoint_anchor_" id="jsPlumb_1_253" style="top: 515px; left: 496.5px;"><a class="task-name" data-toggle="tooltip" data-placement="right" title="Loop over service instance status and fetch docker container logs"><img src="/automation-dashboard/styles/patterns/img/wf-icons/docker.png" width="20px">&nbsp;<span class="name">Loop_Over_Instance_Status</span></a></div><svg style="position:absolute;left:536.5px;top:426.5px" width="51.5" height="89" pointer-events="none" position="absolute" version="1.1"
```
fviale: DOCKER_COMMAND (OPTIONAL)"> this looks like a syntax error. CDATA section is not closed  ]]>

fviale: ok this is the visualization tag, so it's normal, please ignore my comment
.

PR #588, opened 2020-02-18
Update OpenstackInfrastructure and AzureInfrastructure parameters

tao-xinxiu:
Update the nodesource examples based on the [latest change](https://github.com/ow2-proactive/nodesource-addons/pull/80) on OpenstackInfrastructure and AzureInfrastructure parameters 
- OpenstackInfrastructure: downloadCommand -> nodeJarURL
- AzureInfrastructure: downloadCommand -> nodeJarURL; rmHttpUrl -> rmHostname

### DIFF
--- a/NodeSources/resources/catalog/AzureCloud.json
+++ b/NodeSources/resources/catalog/AzureCloud.json
@@ -7,7 +7,7 @@
     "configurableFields": [
       {
         "name": "clientId",
-        "value": "[please put your Azure credentials]",
+        "value": "",
         "meta": {
           "type": "NONE",
           "description": "The Azure clientId",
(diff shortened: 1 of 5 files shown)


NodeSources/resources/catalog/AzureCloud.json
```diff
@@ -226,14 +226,14 @@
         }
       },
       {
-        "name": "downloadCommand",
-        "value": "",
+        "name": "nodeJarURL",
+        "value": "http://[please put the address of the server]:8080/rest/node.jar",
```
amouhoub: Either put a default value (like http://localhost:8080....)
or leave it blank and put some details in the description.
Otherwise, it could be misleading to put a non-working value as a default one, users may consider it as already filled while it's not in reality.

tao-xinxiu: Good point. I remove all the not-working values in default value, and put `try.activeeon.com` as default server address value as other cloud nodesource examples.

PR #610, opened 2020-04-07
Improve PCA instance name unicity check

mklkun:
This PR makes sure that a PCA service instance has a unique name. This guarantees us to have a unique token and a unique PCA instance per Docker container. Some PCA workflows (such as Visdom or Tensorflow) have not been modified yet.

### DIFF
--- a/PcaScripts/resources/catalog/Clean_Start_Service.groovy
+++ b/PcaScripts/resources/catalog/Clean_Start_Service.groovy
@@ -1,0 +1,32 @@
+/*********************************************************************************
+ * THIS CLEANCRIPT MAKE SURE THAT THE PCA INSTANCE STATUS IS IN ERROR IN CASE THE *
+ * SERVICE HAS NOT STARTED PROPERLY                                               *
+ *********************************************************************************/
+
+import org.ow2.proactive.pca.service.client.ApiClient
+import org.ow2.proactive.pca.service.client.api.ServiceInstanceRestApi
+import org.ow2.proactive.pca.service.client.model.ServiceInstanceData
+
+// Acquire variables
+def instanceId = variables.get("PCA_INSTANCE_ID") as long
+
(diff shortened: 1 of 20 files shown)


PcaScripts/resources/catalog/Clean_Start_Service.groovy
```diff
@@ -0,0 +1,32 @@
...
+// Update service instance model (add Deployment, Groups)
+def serviceInstanceData = serviceInstanceRestApi.getServiceInstanceUsingGET(instanceId)
+
+// If service instance is NOT RUNNING then it should be in ERROR
+def currentStatus = serviceInstanceData.getInstanceStatus()
+if (!currentStatus.equals("RUNNING")) {
+    serviceInstanceData.setInstanceStatus("ERROR")
+    serviceInstanceData = serviceInstanceRestApi.updateServiceInstanceUsingPUT(instanceId, serviceInstanceData)
+    println(serviceInstanceData)
+}
```
nebilbenmabrouk, line 32: This section can be enhanced to be more reliable:
In the Start task, if the instance name already exists, write the error state in a text file before throwing an error
In the clean script read the state in the written file, if it is error, put the service instance in ERROR state

mklkun: Yes, I agree,  it should improve this scenario.  I will fix that.

mklkun: As I told you in Slack. Upon reflection, I think it's better to leave this test as it is. This being for 2 reasons:
- It is a clean script, which will ensure that the state of the instance is in `ERROR` if after the Start task (for some reason or another) it is not in a `RUNNING` state.
- To avoid that the state remains in transition indefinitely.
- To have a clean script as independent as possible from the launch script

PR #612, opened 2020-04-10
Fix docker container search by name

mklkun:


CloudAutomationDockerTemplate/resources/catalog/PCA_Template.xml
```diff
@@ -102,15 +102,15 @@ if [ -z "$INSTANCE_NAME" ]; then
     exit 1
 fi
 
-if [ "$(docker ps --format '{{.Names}}' | grep "^$INSTANCE_NAME$")" ]; then
+if [ "$(docker ps -a --format '{{.Names}}' | grep "^$INSTANCE_NAME$")" ]; then
```
mcompastie, line 105: Hum ... docker ps -a will show also unloaded containers. Should you assess it is running before stating it is in use by another service instance ?
This comment apply to all modified WF.

mklkun: Even if stopped/paused, it already exists and it is ready to be resumed as in our main pipeline when finished we use a `docker rm`. To avoid issues we should not have two PCA service instances with the same name to avoid issues between the containers and tokens management.

mklkun: Plus, I just checked, even if a docker container is stopped/exited, we can't run another one having the same name.
`docker: Error response from daemon: Conflict. The container name "/myJupyterHub" is already in use by container "7371f090c6ae3cec8657d9a6cfe02210f99291f5e7792afd7bfdd5258d110653". You have to remove (or rename) that container to be able to reuse that name.
See 'docker run --help'.`

mcompastie: I see. In that case, it is all good to me.

PR #627, opened 2020-04-29
Add MESSAGE as a workflow variable (required to submit it from the python connector)

andrewssobral:


WebNotification/resources/catalog/web_notification.xml
```diff
@@ -21,7 +24,7 @@
         <![CDATA[ Task to send a message to the notification service ]]>
       </description>
       <variables>
-        <variable name="MESSAGE" value="Put your message to the notification service here" inherited="false" />
+        <variable name="MESSAGE" value="Put your message to the notification service here" inherited="true" />
       </variables>
```
aminelouati, line 28: If the variable is inherited no need to put the same value.

andrewssobral: No problem, i would like to keep the same behavior

PR #639, opened 2020-05-13
Fixed MaaS state transitions, added Update_Model_Service pca-workflow

andrewssobral:
image


andrewssobral:
Thanks a lot @aminelouati @ShatalovYaroslav

PR #644, opened 2020-05-15
Fix timeout in PCA Service_Action script

mklkun:


PcaScripts/resources/catalog/Service_Action.groovy
```diff
@@ -71,7 +73,11 @@ if( !actionVariables.isEmpty() ){
 def serviceInstanceData = serviceInstanceRestApi.launchServiceInstanceActionUsingPUT(sessionId, instanceId as int, service, variables.get("PA_JOB_ID"))
 
 if (action.toLowerCase().contains("finish")) {
-    schedulerapi.waitForJob(serviceInstanceData.getJobSubmissions().get(0).getJobId().toString(), 30000)
+    try {
+        schedulerapi.waitForJob(serviceInstanceData.getJobSubmissions().get(0).getJobId().toString(), 180000)
```
mcompastie, line 77: Please assess at least one element in service instance data exist. Otherwise, we could go through NullPointerException :-/

mklkun: It is a script meant to be executed in a user workflow. It is impossible that `JobSubmissions` is null if the target PCA service is up. If it is not, the task should be faulty in any case because we can't execute an action on a PCA service that does not exist.

PR #669, opened 2020-05-29
 Fix email validation workflow

fviale:
- make the email validation workflow compatible with new notification-service
 - remove jython usage in docker_exec_env and PostgreSQL workflows


Email/resources/catalog/email_validation.xml
```diff
@@ -1,74 +1,136 @@
...
+def paSchedulerRestUrl = variables.get('PA_SCHEDULER_REST_URL')
+def notifUrl = paSchedulerRestUrl.replaceAll("/rest\\z", "/notification-service")
+// Instantiate ValidationRestApi instance
+def apiClient = new ApiClient()
+apiClient.setBasePath(notifUrl)
+def validationRestApi = new ValidationRestApi(apiClient)
+//Get job id
+def jobId = new Long(variables.get("PA_JOB_ID"))
+println(jobId)
+println(variables.get("PA_JOB_ID"))
```
AlexFalessi, line 54: I wonder if displaying those fields is wanted ?

fviale: I think so, you receive a validation email from a specific job. The goal of the validation is to unlock this job.

PR #696, opened 2020-07-22
Fix and update Send_Result_By_Email:

AlexFalessi:
- Fix case where only logs should be in email
- If there are no results, email is not sent
- Code optimizations


Email/resources/catalog/Send_Results_By_Email.xml
```diff
@@ -76,99 +83,97 @@ def tasksLogsFile = new File("job"+jobId+"-logs.log");
...
-        	resultFile.createNewFile();
-    		resultFile.append(item.getValue().toString());
-			resultFiles.add(resultFile);
-        }
-        
-        addLogs(item, addTasksLogs, tasksLogsFile)
+if (addTasksResults || addTasksLogs) {
+
+    // If the first element of the results is the default "true" string, there are no results
+    if (!results[0].toString().equals("true")) {
```
fviale, line 104: please do :
`if (results[0] != null && !results[0].toString().equals("true")) {`

PR #700, opened 2020-08-13
fix: update cloud node source examples to add parameter startupScript

tao-xinxiu:
Add parameter startupScript and update some fields description and default value, to make the node source examples to match their updated infrastructure fields definition.

### DIFF
--- a/NodeSources/resources/catalog/AzureCloud.json
+++ b/NodeSources/resources/catalog/AzureCloud.json
@@ -56,7 +56,7 @@
           "type": "NONE",
           "description": "Optional authentication endpoint from specific Azure environment",
           "dynamic": false,
-          "sectionSelector": 2,
+          "sectionSelector": 3,
           "important": false
         }
       },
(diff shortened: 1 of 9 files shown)


NodeSources/resources/catalog/AzureCloud.json
```diff
@@ -265,20 +265,123 @@
...
+          "sectionSelector": 2,
+          "important": false
+        }
+      },
+      {
+        "name": "locale",
+        "value": "en-US",
+        "meta": {
+          "type": "NONE",
+          "description": "The culture in which the resource metadata needs to be localized.",
```
MaelAudren, line 321: `culture` seems weird to me, maybe `language code` or `language localization` would be better.

tao-xinxiu: I agree. But this description actually is taken from nodesource-addons project AzureInfrastructure field description. I think it would be better to change it in the next version.

PR #741, opened 2020-09-04
Adapt MongoDB to support singularity

ImenBizid:
In this PR, we:
- adapt mongo PCA  to support singularity
- add the required variables to make mongo PCA work for both singularity and Docker
- delete the Resume and Pause actions


MongoDB/resources/catalog/Finish_MongoDB.xml
```diff
@@ -1,25 +1,162 @@
...
+        <info name="task.icon" value="/automation-dashboard/styles/patterns/img/wf-icons/mongodb.png"/>
+        <info name="Documentation" value="PCA/PCAUserGuide.html"/>
+        <info name="NODE_ACCESS_TOKEN" value="$INSTANCE_NAME"/>
+      </genericInformation>
+      <pre>
+        <script>
+          <code language="bash">
+            <![CDATA[
+CLEAN_DATA=$variables_CLEAN_DATA
+if [ "$CLEAN_DATA" = "true" ]; then
```
fviale, line 32: I believe here it should be ==

andrewssobral: Yes me too, and also using Bash4 lowercase mode ${CLEAN_DATA,,}

andrewssobral:
@fviale I asked to @ImenBizid to put this PR on hold until I finish all tests on HPE (SLURM+Singularity), surely new fixes will be done after that.

MongoDB/resources/catalog/Finish_MongoDB.xml
ImenBizid: I fixed it. Thanks for the remark

andrewssobral (approved):

andrewssobral:
@ImenBizid @fviale I'm approving this PR because it's working fine for Docker + AutoML, but not with Singularity. It seems to be a Dockerfile or maybe some Singularity parameters that should be checked. If we need to change something inside the MongoDB PCA service I will create another PR.

andrewssobral:
@ImenBizid @fviale I'm approving this PR because it's working fine for Docker + AutoML. However, with Singularity, the service is UP and RUNNING, but we got Authentication Failed. This is caused because the entrypoint file is not executed after docker-to-singularity image conversion. It seems to be a Dockerfile or maybe some Singularity parameters that should be checked. If we need to change something inside the MongoDB PCA service I will create another PR.

PR #752, opened 2020-09-30
Fix token use instead of password due to websocket issue

mklkun:


JupyterLab/resources/catalog/JupyterLab.xml
```diff
@@ -10,7 +10,7 @@
     <variable name="PROXYFIED" value="False" model="PA:Boolean"/>
     <variable name="HTTPS_ENABLED" value="False" model="PA:Boolean"/>
     <variable name="SERVICE_PORT" value="-1" model="PA:Integer"/>
-    <variable name="JUPYTERLAB_PASSWORD" value="" model="PA:HIDDEN"/>
+    <variable name="JUPYTERLAB_TOKEN" value="" model="PA:HIDDEN"/>
```
fviale, line 13: is it necessary to change the variable name ? from thr user point of view, I think password is more understanable

mklkun: No it is not necessary. I thought that it can be a confusion when the jupyter server asks for a token and the user has set a password.

fviale: ok

fviale (approved):

PR #777, opened 2020-11-30
Docker fork environment: do not mount volume when a specific java property is defined

fviale:
Also, add shm-size parameter in some fork env scripts


Scripts/resources/catalog/fork_env_cuda_universal.groovy
```diff
@@ -245,7 +261,7 @@ if (CONTAINER_ENABLED && (
     }
 
     // Prepare log directory
-    if (HOST_LOG_PATH && CONTAINER_LOG_PATH) {
+    if (!isPANodeInContainer && HOST_LOG_PATH && CONTAINER_LOG_PATH) {
```
mboussaa, line 264: logs won't be able to be fetshed in `CONTAINER_LOG_PATH` as they are not part of `volumes-from`. I don't know if this matters

mboussaa (approved):

Scripts/resources/catalog/fork_env_cuda_universal.groovy
fviale: Most often the default values of CONTAINER_LOG_PATH will be /shared/something. So if the container mounts /shared it should be ok. In other cases, it will not work indeed

PR #778, opened 2020-12-08
Adapt Remote Visualization workflow to new API

fviale:


aminelouati:
Can this workflow example be automatically tested via workflow-test ? If yes, you need to add it scenario.json file.

aminelouati (approved):

tao-xinxiu (approved):

fviale:
@aminelouati I guess it is already.

fviale:
@but wait, the damn workflow is called "remote_visulalization", I have to do same renaming !

PR #789, opened 2021-01-15
Add a variable with a PA:Credential model to ensure that DB credentials are added before submission

aminelouati:


DataConnectors/resources/catalog/SQL_Pooled_Connection_Query.xml
```diff
@@ -3,10 +3,11 @@
   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
   xmlns="urn:proactive:jobdescriptor:3.11" xsi:schemaLocation="urn:proactive:jobdescriptor:3.11 http://www.activeeon.com/public_content/schemas/proactive/jobdescriptor/3.11/schedulerjob.xsd"  name="SQL_Pooled_Connection_Query" projectName="2. SQL" priority="normal" onTaskError="continueJobExecution"  maxNumberOfExecution="2" >
   <variables>
-    <variable name="RDBMS_NAME" value="HSQLDB" model="PA:LIST(PostgreSQL, MySQL, Greenplum, SQL Server, Oracle, HSQLDB)"/>
+    <variable name="RDBMS_NAME" value="hsqldb" model="PA:LIST(postgresql, mysql, greenplum, sqlserver, oracle, hsqldb)"/>
     <variable name="HOST" value="localhost" />
-    <variable name="PORT" value="" />
+    <variable name="PORT" value="9001" />
```
fviale, line 8: I think it should be left empty, no ?

aminelouati: Better no because it is a part of the cred key

DataConnectors/resources/catalog/SQL_Pooled_Connection_Query.xml
```diff
@@ -123,12 +124,9 @@ if (!database) {
...
 // This key is used for getting the password from 3rd party credentials.
-CREDENTIALS_KEY = RDBMS_PROTOCOL + "://" + username + "@" + host + ":" + port
-CREDENTIALS_KEY_MSG = RDBMS_PROTOCOL + "://<username>@<host>:<port>"
-password = credentials.get(CREDENTIALS_KEY)
-if (!password) {
-    println("WARNING: Your didn't add your password to 3rd-party credentials under the key :\"" + CREDENTIALS_KEY + "\"")
-}
+CREDENTIALS_KEY = RDBMS_NAME + "://" + username + "@" + host + ":" + port
+CREDENTIALS_KEY_MSG = RDBMS_NAME + "://<username>@<host>:<port>"
+password = credentials.get(CREDENTIALS_KEY).trim()
```
aminelouati, line 129: I used trim() here to consider the passwordless auth

DataConnectors/resources/catalog/SQL_Pooled_Connection_Query.xml
fviale: but I thought the port is determined dynamically based on the RDMS

fviale: why not remove the port from the credential key ? 
It is not necessary to have the port

fviale: additionnally, for the port, you can use model PA:INTEGER?

aminelouati: OK to remove the port from the credential key
If we add PA:Integer model --> the default value is mandatory

fviale: not by adding a question mark at the end : optional model

fviale (approved):
LGTM, thanks amine

PR #790, opened 2021-01-18
Add ETL and PySpark wkws

mbenguig:


GetStarted/METADATA.json
```diff
@@ -126,12 +126,31 @@
...
+					"contentType": "application/xml"
+				},
+				"file" : "resources/catalog/iterative_etl.xml"
 			}
 		]
 	},
 	"dataspace" : {
 		"target" : "global",
 		"files" : ["resources/dataspace/canny-edge-detector.jar",
+            "resources/dataspace/mysql-connector-java-8.0.22.jar",
```
aminelouati, line 153: I don't think that this jar is needed as you are using the python template of mysql connector

GetStarted/METADATA.json
```diff
@@ -126,12 +126,31 @@
...
+				"file" : "resources/catalog/multi_source_etl.xml"
+			},
+			{
+				"name" : "Iterative_ETL",
+				"metadata" : {
+					"kind": "Workflow/standard",
+					"commitMessage": "First commit",
+					"contentType": "application/xml"
+				},
+				"file" : "resources/catalog/iterative_etl.xml"
```
aminelouati, line 146: I think that we have to think about a better organization for the basic-examples bucket. For me the current name does not reflect what it contains.

aminelouati (approved):

GetStarted/METADATA.json
mbenguig: I would say yes since I needed it .. but I will check again!

GetStarted/METADATA.json
mbenguig: You're right ... I suggested this bucket since it contains the "advanced workflow" project. Ok find a way to centralize etl workflows into a dedicated bucket.

GetStarted/METADATA.json
mbenguig: It is required by the scala/spark task, at the spark script submission level:

def spark_shell_command = "/usr/local/spark/bin/spark-shell --driver-memory 800m --executor-memory 800m --master spark://" + instance_name + "-spark-master:7077 --jars /usr/local/spark/jars/*,mysql-connector-java-8.0.22.jar -I script.scala --conf spark.driver.args='" + mysql_host + " " + mysql_port + " " + mysql_user + " " + mysql_passwd + " " + mysql_database + " " + mysql_table + " " + hdfs_master + " " + output_hdfs_csv_file_name + "'"

GetStarted/METADATA.json
mbenguig: following our discussion (some ML wkws could be considered as ETL wkws)...  I will leave it like this.

PR #793, opened 2021-01-19
Add clean token script in finish PCA services action

mklkun:


aminelouati (commented):
I am not sure but I think that you need to add this clean mechanism not only to the Finish action but also to all the other actions. 
What we observe actually is when an issue occur at the docker level (in the loop task) and the service fails, the token is not removed.

mklkun:
> I am not sure but I think that you need to add this clean mechanism not only to the Finish action but also to all the other actions.
> What we observe actually is when an issue occur at the docker level (in the loop task) and the service fails, the token is not removed.

I don't think Amine. When other actions are triggered (mainly pausing or resuming), the PCA service will still be not finished n the Cloud Automation microservice. This gives the right to users to trigger other actions (resuming, pausing finishing) from the portal. These actions need the token to be present on the node to be executed. 
The only situation we need to clean the token is when the main workflow ended prematurely for some reason, and the user executed the finish action.

aminelouati:
Sorry Khalil I was wrong.
However, in case an issue occurs at the docker level, do we have a mechanism to clean the token when we kill manually the workflow ?

aminelouati:
test

mklkun:
> Sorry Khalil I was wrong.
> However, in case an issue occurs at the docker level, do we have a mechanism to clean the token when we kill manually the workflow ?

Yes. In this case the loop that is implemented checks if the docker container is still running. If not, the PCA service is set to be in an `ERROR` state. In this case when the user kills the service, the token will be cleaned.

aminelouati (approved):
LGTM

PcaScripts/resources/catalog/Clean_Trigger_Action.groovy
```diff
@@ -0,0 +1,43 @@
...
+def sessionId = schedulerapi.getSession()
+
+// Connect to Cloud Automation API
+def apiClient = new ApiClient()
+apiClient.setBasePath(pcaUrl)
+def serviceInstanceRestApi = new ServiceInstanceRestApi(apiClient)
+
+// Update service instance model (add Deployment, Groups)
+def serviceInstanceData = serviceInstanceRestApi.getServiceInstanceUsingGET(sessionId, instanceId)
+serviceInstanceData = serviceInstanceRestApi.getServiceInstanceUsingGET(sessionId, instanceId)
```
AlexFalessi, line 32: Is the same rest call twice wanted ?

mklkun: Nope. You're right. I forgot to delete the second line. Nice find!

AlexFalessi (approved):

PR #799, opened 2021-02-02
Update signal control tasks

nebilbenmabrouk:


Controls/METADATA.json
```diff
@@ -110,22 +110,22 @@
...
 				"metadata" : {
 					"kind": "Workflow/standard",
 					"commitMessage": "First commit",
 					"contentType": "application/xml"
 				},
-				"file" : "resources/catalog/Wait_For_Signal.xml"
+				"file" : "resources/catalog/Wait_For_Signals.xml"
 			}
 		]
 	}
```
aminelouati, line 131: Can these workflows be automatically tested. If yes can you add them to workflow-test ?
Thanks

aminelouati (approved):

Controls/METADATA.json
nebilbenmabrouk: No these workflows cannot be tested automatically because they execute/wait continuously until a signal is received. They do not stop executing by themselves.

Controls/resources/catalog/Check_For_Signals.xml
```diff
@@ -0,0 +1,154 @@
...
+signalsList =  variables.get("SIGNALS_LIST")
+
+// Check whether one signal among those specified as input is received
+println("Checking whether one signal in the list "+ signalsList +" is received")
+receivedSignal = signalapi.checkForSignals(signalsList)
+
+// If a signal is received, break the loop, else sleep 10 seconds and restart
+if (receivedSignal != null){
+    println("Received signal: "+ receivedSignal)
+    variables.put("RECEIVED",true)
```
fviale, line 100: you could also here set the received signal as result

fviale (approved):
two minor comments, otherwise LGTM

PR #814, opened 2021-03-04
Adding some features to the DDD (update, notification, removing old DDD)

dianajlailaty:
In this PR, the following was applied:

1. adding the update feature to DDD in PSA MaaS_ML and Swagger.
2. adding the drift notification to DDD.
3. removing drift threshold and model metadata from Studio and Swagger.


andrewssobral (approved):
Approved! Thanks @dianajlailaty

ImenBizid (approved):

PR #838, opened 2021-03-18
Create PSA services management workflows

mklkun:


mklkun:
retest this please

marounkoussaifi (approved):

COCOAnnotator/METADATA.json
```diff
@@ -31,6 +31,15 @@
           "contentType": "application/xml"
         },
         "file": "resources/catalog/Finish_COCO_Annotator.xml"
+      },
+      {
+        "name" : "COCO_Annotator_Service",
```
nebilbenmabrouk, line 36: The naming convention `xxx_Service` is not clear to me and can be confused with the PSA wf/service itself.
I prefer  `xxx_Launcher` or  `xxx_Management_Worrkflow` unless this suggestion comes from Denis :)

COCOAnnotator/resources/catalog/COCO_Annotator_Service.xml
```diff
@@ -0,0 +1,191 @@
...
+// Wait until one signal among those specified is received
+println("Waiting for any signal among "+ signalsSet)
+receivedSignal = signalapi.waitForAny(signalsSet)
+
+// Remove ready signals
+signalapi.removeManySignals(new HashSet<>(signalsSet.collect { signal -> "ready_"+signal }))
+
+// Display the received signal and add it to the job result
+println("Received signal: "+ receivedSignal)
+result = receivedSignal
```
nebilbenmabrouk, line 155: If the same code will be used by all services, it is better to create a script and reference it.

COCOAnnotator/METADATA.json
mklkun: I understand your point. But this was discussed and decided by Denis, so ... 
In any case I think it is ok to have `xxx_Service` since these workflows will be seen alone in `service-automation` from workflow-execution.

nebilbenmabrouk (approved):
LGTM

PR #841, opened 2021-03-18
Update jre

alijawadfahs:


JobAnalysis/resources/catalog/Gantt_chart.xml
```diff
@@ -30,7 +30,7 @@
       <genericInformation>
         <info name="task.icon" value="/automation-dashboard/styles/patterns/img/wf-icons/gantt_chart_ter.png"/>
       </genericInformation>
-      <forkEnvironment javaHome="/root/jre1.8.0_131" >
+      <forkEnvironment javaHome="/root/jre1.8.0_281" >
```
fviale, line 33: This one is a bit weird. I believe the instruction is superseded by the content of the fork environment script fork_env_docker_args  on Linux (as the script mounts the host jre inside the containter).
On windows or Mac, I don't know if it works as the container activeeon/pa_plot does not seem to contain the jre.

fviale (approved):
one comment, otherwise LGTM

JobAnalysis/resources/catalog/Gantt_chart.xml
mboussaa: @fviale @alijawadfahs I agree, only jre1.8.0_131 exist
```
root@588cca44bc63:~# pwd
/root
root@588cca44bc63:~# ls -la
total 28
drwx------ 1 root root 4096 Sep 27  2017 .
drwxr-xr-x 1 root root 4096 Mar 21 22:07 ..
-rw-r--r-- 1 root root  681 Sep 27  2017 .bashrc
drwxr-xr-x 3 root root 4096 Sep 27  2017 .conda
-rw-r--r-- 1 root root  140 Nov 19  2007 .profile
drwxr-xr-x 6 uucp  143 4096 Mar 15  2017 jre1.8.0_131
drwxr-x--- 9 1000 1002 4096 Sep 27  2017 pa_plot
root@588cca44bc63:~#
```
It won't work on windows as the JRE mount is not done for WS. I think we can keep /root/jre1.8.0_131 for this case or add 0_281 in docker image if we want to integrate this change

alijawadfahs: I guess let's upgrade it for consistency

fviale: I would say let's not bother. I don't know how to recreate the docker container.

mboussaa (approved):

JobAnalysis/resources/catalog/Gantt_chart.xml
mboussaa: Ok lets keep 131 for this property

PR #892, opened 2021-04-20
Add HDInsight Spark workflows

nebilbenmabrouk:


aminelouati (changes requested):
Please add PA:model for workflow variables such as PA:HIDDEN, PA:NOT_EMPTY_STRING, etc as much as possible especially for password variables (e.g. CLUSTER_PASSWORD) to control and improve the workflow execution.

aminelouati (approved):

PR #903, opened 2021-05-04
change protocol names of PSA services

aminelouati:


Greenplum/resources/catalog/Greenplum.xml
```diff
@@ -196,7 +196,7 @@ echo "$containerID" > $INSTANCE_NAME"_containerID"
         <script>
           <file url="${PA_CATALOG_REST_URL}/buckets/service-automation/resources/Post_Start_Service/raw" language="groovy">
             <arguments>
-              <argument value="tcp"/>
+              <argument value="greenplum"/>
```
fviale, line 199: I think greenplum is a postgresql database
https://docs.greenplum.org/6-4/pxf/jdbc_cfg.html

Logstash/resources/catalog/Logstash.xml
```diff
@@ -104,7 +104,7 @@ echo END "$variables_PA_TASK_NAME"
         <script>
           <file url="${PA_CATALOG_REST_URL}/buckets/service-automation/resources/Post_Start_Service/raw" language="groovy">
             <arguments>
-              <argument value="tcp"/>
+              <argument value="logstash"/>
```
fviale, line 107: I am not sure there is a logstash protocol. The following refers to mysql
https://www.elastic.co/guide/en/logstash/current/plugins-inputs-jdbc.html

Please contact the workflow designer to clarify

Zookeeper/resources/catalog/Zookeeper.xml
```diff
@@ -101,7 +101,7 @@ echo END "$variables_PA_TASK_NAME"
         <script>
           <file url="${PA_CATALOG_REST_URL}/buckets/service-automation/resources/Post_Start_Service/raw" language="groovy">
             <arguments>
-              <argument value="tcp"/>
+              <argument value="zookeeper"/>
```
fviale, line 104: again, it is not clear what the zookeeper endpoint is (http, etc), please contact the designer

amouhoub:
@aminelouati @fviale For RMDBs (oracle, postgres, mysql, etc) the connection URL depends on the software that would connect to it. For example from a java app using JDBC it would be `jdbc:postgres://hostna....`. From another c/c++ software on windows it would be odbc....
It's not correct either to have for example : `postgres//hostna...` as an endpoint for Postgres or Greenplum services as they simply don't provide such a connection url.
Mongo has its own protocol because they have their own sdk for multiple languages but it would be an exception.
So normally if a docker image doesn't provide an Open in browser button inside the Docker Desktop app, then we shouldn't publish it.
The aim of providing an endpoint for a PSA service should be either to be reused from another job using the exact same endpoint url & protocol, or to be opened in the browser from the portal.
If it's neither the case then we simply shouldn't publish any endpoint instead of publishing a wrong one.
Connection info for a RDBMS should not be provided to the users in the portal through a copy-paste WRONG link, even from a UX point of view he will need to copy its small bits one by one (hostname, port, etc) in his own software.
We need stop creating false endpoints and find a better way to display the inner service info (not the PSA service info)...
So in your PR @aminelouati I suggest that you simply leave the argument blank so no endpoint will be published
We can have a discussion about that if you feel up to it @fviale @aminelouati @mboussaa

fviale:
@amouhoub 

Yes,
"The aim of providing an endpoint for a PSA service should be either to be reused from another job using the exact same endpoint url & protocol, or to be opened in the browser from the portal.", 
So let's provide the endpoint uri as we would use it in a workflow.

Zookeeper/resources/catalog/Zookeeper.xml
aminelouati: Normally the argument has to be removed. However, if I remove it, zookeeper service will not create an endpoint which is required by Kafka to retrieve the host and port information. The implementation of Kafka service has to be changed.
For now I will keep zookeeper unless you want me to change it to something else.

aminelouati: The idea of the argument if provided is to precise the protocol of the service. Otherwise, the service will not expose any endpoint.

fviale (approved):

mklkun (approved):

PR #904, opened 2021-05-04
Fix is mac

mboussaa:


mboussaa:
@andrewssobral Thank you for pointing out the issue. Could you review it?

andrewssobral:
@mboussaa I think the best is to move out isWindows and isMac setting the default value as False, to avoid missing variable definition.

do this instead:
```
isWindows = false
isMac = false
switch (family) {
        case OperatingSystemFamily.WINDOWS:
            isWindows = true
            break
        case OperatingSystemFamily.MAC:
            isMac = true
            break
    }
```

Note: groovy scripts does not need `;` at the end of line

mboussaa:
@andrewssobral done

andrewssobral (approved):

andrewssobral:
LGTM! let's merge it ;-)

PR #947, opened 2021-06-03
Improve remote visualization by generating the number of display and ports randomly

nebilbenmabrouk:
…ports randomly


HPC/resources/catalog/Gromacs_VMD_Molecule_Simulation.xml
```diff
@@ -59,13 +57,28 @@ The workflow takes as input:
  * visu_command : command used to start the VNC session (e.g. "Xvnc :${DISPLAY} -geometry 1280x1024 -SecurityTypes None")
  */
 
-def display = args[0]
-def visu_command = args[1]
+def visu_command = args[0]
+findPortScript = new File("find_port.sh")
```
fviale, line 61: could you please delete the script after being used.

Also, it would be clearer if the findPort code is put inside a function

nebilbenmabrouk: Done @fviale

fviale (approved):
LGTM

HPC/resources/catalog/Gromacs_VMD_Molecule_Simulation.xml
```diff
@@ -59,13 +57,37 @@ The workflow takes as input:
  * visu_command : command used to start the VNC session (e.g. "Xvnc :${DISPLAY} -geometry 1280x1024 -SecurityTypes None")
  */
 
-def display = args[0]
-def visu_command = args[1]
+// Generate random port for remote visualization
+def findPort() {
+    findPortScript = new File("find_port.sh")
```
AlexFalessi, line 62: Shouldn't the file creation + write + delete be in a try catch block to handle IOException?

nebilbenmabrouk: @AlexFalessi There is no specific behaviour to add in the catch block. If an exception occurs, it will be thrown with an explicit message.

AlexFalessi (approved):

PR #955, opened 2021-06-29
Adapt some PSA services to be able to run in a virtualized environment

aminelouati:


JupyterLab/resources/catalog/JupyterLab.xml
```diff
@@ -7,7 +7,7 @@
     <variable name="NATIVE_SCHEDULER_PARAMS" value=""/>
     <variable name="INSTANCE_NAME" value="jupyterlab-server-$PA_JOB_ID"/>
     <variable name="ENDPOINT_ID" value="jupyterlab-endpoint-$PA_JOB_ID"/>
-    <variable name="PROXYFIED" value="False" model="PA:Boolean"/>
+    <variable name="PROXYFIED" value="True" model="PA:Boolean"/>
```
mklkun, line 10: Why do we have to switch the default value to `True`? Is this mandatory?

fviale: I made this request :-)

It is mandatory in proactive server deployed in https due to the http header strict-transport-security 
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security

JupyterLab/resources/catalog/Jupyterlab_Service.xml
```diff
@@ -6,7 +6,7 @@
     <variable name="NATIVE_SCHEDULER" value=""/>
     <variable name="NATIVE_SCHEDULER_PARAMS" value=""/>
     <variable name="ENDPOINT_ID" value="jupyterlab-endpoint-$PA_JOB_ID"/>
-    <variable name="PROXYFIED" value="False" model="PA:Boolean"/>
+    <variable name="PROXYFIED" value="True" model="PA:Boolean"/>
```
fviale, line 9: can you keep it lowercase ?

JupyterLab/resources/catalog/JupyterLab.xml
fviale: @aminelouati There must be a default password configured for JupyterLab (which should be explained in the workflow description)

mklkun: Fair enough. But then a second question, why not changing this default value for all our PSA services?

JupyterLab/resources/catalog/Jupyterlab_Service.xml
mklkun: This is seen just in the xml file. It was used with capital F/T in all PSA services. I agree that in lowercase it is better but we should also do this change in all PSA service workflows (management workflows will be deleted anyway :'( lol)

JupyterLab/resources/catalog/JupyterLab.xml
fviale: It should be done on all PSA service which currently support proxyfied mode.

aminelouati: @fviale OK I will add a default password for JupyterLab explained in the workflow description.
But for info even with an empty password, the jupyterLab web interface is accessible in a proxyfied mode.

aminelouati: I mean we can login with an empty password

fviale: @aminelouati So it displays the form to enter a password and you enter nothing inside ?

aminelouati: Yes exactly. 
Another point, some services such as Knime are not working anymore even with the proxyfied mode. When I opened network tab in the developer tools I got this error
Status Code: 504 Gateway Timeout

fviale: Regarding the password, it is ok to leave it empty, but the description should include this mention.

I don't think work as been done on Knime regarding the proxyfied mode.

aminelouati: As you said knime and many other services do not support the proxified mode. 
I will create a separate card to update them.

PR #959, opened 2021-07-07
Update Kubernetes node source example

tao-xinxiu:


NodeSources/resources/catalog/K8SClusterElastic.json
```diff
@@ -176,7 +198,7 @@
       },
       {
         "name": "schedulerUrl",
-        "value": "pnp://[please put the address of the server]:64738/",
+        "value": "pamr://4096",
```
fviale, line 201: not sure all these settings should be changed. The node source template should not use the try platform configuration, but be generic

tao-xinxiu: Some discussion about this were made before on the other node source examples, since users often not checking the value like "[please put the address of the server]", and only check them when encountering an error. So it might be better to either leave it empty (e.g. for credentials, to trigger the error, force user to change) or put some real example. 
Most node sources has put try platform config for the related parameters (e.g., rmHostname, nodeJarURL etc). So I've kept it coherent with the others.

fviale: Ok, but I preferred the other approach, as here it is unclear for the user what he needs to change.

If it has been done like this for other node sources, so be it, but I don't like it.

fviale (approved):

alijawadfahs (approved):

PR #979, opened 2021-08-27
Add userGroup attribute to all WFs

marounkoussaifi:


fviale:
I don't see a problem with this PR, but I am not sure it brings anything. We are not supposed to define groups in the catalog as groups definition depends on the customer

fviale:
are all objects belonging to pubilc-objects or are there exceptions ?

fviale:
@marounkoussaifi ok I get it, thanks

fviale (approved):
LGTM

PR #1002, opened 2021-11-05
refactor-distributed-automl-workflow

oanaBiancaSchiopu:


fviale (approved):

AlexFalessi (commented):
I think you can use Boolean class instead of doing lower case and compare.
`Boolean.valueOf(variables['VISDOM_ENABLED'])? showVar('VISDOM_PROXYFIED') : hideVar('VISDOM_PROXYFIED')`

PR #1008, opened 2021-11-19
improve some workflows of the controls bucket

aminelouati:


mboussaa (approved):

oanaBiancaSchiopu (approved):

fviale (changes requested):
I disagree with this change. The Controls menu does not contain workflows which are meant to be executed direcly like basic_examples. It contains templates which must be dragged /dropped when building a workflow

PR #1013, opened 2021-11-26
move etl workflows to the BigData bucket

fviale:


aminelouati (approved):
Could you remove the following GI from the Iterative_ETL workflow or change the value to big-data ? 
<info name="bucketName" value="basic-examples"/>

PR #1027, opened 2021-12-09
change-check-signals-return-type

oanaBiancaSchiopu:


Controls/resources/catalog/Check_For_Signals.xml
```diff
@@ -92,17 +92,17 @@ Set signalsSet =  variables.get("SIGNALS_SET")
...
 
     // remove ready signals
 	signalapi.removeManySignals(new HashSet<>(signalsSet.collect { signal -> "ready_"+signal }))
 
-    // print the received signal
-    println("Received signal: "+ receivedSignal.getName())
-    result = receivedSignal.getName()
+    // print the received signals
+    println("Received signals: "+ receivedSignals.keySet())
+    result = receivedSignals.keySet()
```
nebilbenmabrouk, line 105: At this point we need to check/change the code of the loop accordingly.

Controls/resources/catalog/Check_For_Signals.xml
```diff
@@ -92,17 +92,17 @@ Set signalsSet =  variables.get("SIGNALS_SET")
...
+if (receivedSignals != null && !receivedSignals.isEmpty()){
 
     // remove ready signals
 	signalapi.removeManySignals(new HashSet<>(signalsSet.collect { signal -> "ready_"+signal }))
 
-    // print the received signal
-    println("Received signal: "+ receivedSignal.getName())
-    result = receivedSignal.getName()
+    // print the received signals
+    println("Received signals: "+ receivedSignals.keySet())
```
nebilbenmabrouk, line 104: It is good also to print the values of the signal variables

Controls/resources/catalog/Wait_For_Signals.xml
```diff
@@ -57,17 +57,17 @@ Set signalsSet =  variables.get("SIGNALS_SET")
...
 
     // remove ready signals
 	signalapi.removeManySignals(new HashSet<>(signalsSet.collect { signal -> "ready_"+signal }))
 
-    // print the received signal
-    println("Received signal: "+ receivedSignal.getName())
-    result = receivedSignal.getName()
+    // print the received signals
+    println("Received signals: "+ receivedSignals.keySet())
+    result = receivedSignals.keySet()
```
nebilbenmabrouk, line 70: Same here, it will be good to print the values of the signals' variables

nebilbenmabrouk (commented):
I noticed that the same code of (Check_For_Signals) is repeated and used in all the service workflows.
I suggest to add this code in a script in the signal bucket (we call it 'check_for_signals.groovy'), and then reference it in all the workflows. This is good for modularity and the maintenance of the workflows.

PR #1044, opened 2022-01-03
Add the AutoFeat endpoint

Hiba-Alili:


dianajlailaty (approved):

mklkun (approved):

MachineLearning/resources/catalog/Import_Data_Interactive_Script.py
```diff
@@ -111,8 +111,8 @@ def get_open_port():
 flask_port = get_open_port() # 8001
 bokeh_url = "http://"+external_ip+":"+str(bokeh_port)
 flask_url = "http://"+external_ip+":"+str(flask_port)
-print("bokeh_url: ", bokeh_url)
-print("flask_url: ", flask_url)
+
+schedulerapi.addExternalEndpointUrl(variables.get("PA_JOB_ID"), "AutoFeat", bokeh_url, "/automation-dashboard/styles/patterns/img/wf-icons/data-processing.png")
```
mboussaa, line 115: And flask ?

mboussaa: shouldn't we add schedulerapi.connect() before ?

Hiba-Alili: i already got the access to schedulerapi and acquired a session id before. Please check the line 48.

Hiba-Alili: Actually, we're embedding a bokeh content into a flask web app. The bokeh server should be running in the background to get its content and embed it into the flask app. So we need to provide access only to the flask app, instead.  Thank you for the remark.

mboussaa (approved):

PR #1087, opened 2022-02-04
refactor-wait-for-signals-with-variables

oanaBiancaSchiopu:
Reactor _wait-for-signals-with-variables_ workflow to send signal with variables just for _Continue_ signal


mklkun (approved):
LGTM

fviale (approved):

PR #1136, opened 2022-05-09
[automl] added multi-objective support

carolinepacheco:


mboussaa:
@carolinepacheco I don't see how the "multi-objective support" algorithm is implemented in ur PR. Could you please tell me how can I check it? thx

Hiba-Alili (approved):

andrewssobral:
> @carolinepacheco I don't see how the "multi-objective support" algorithm is implemented in ur PR. Could you please tell me how can I check it? thx

@mboussaa the algorithm exists already (MOCMAES - https://doc.activeeon.com/latest/PML/PMLUserGuide.html#_AutoML ), however the AutoML workflow not supported multi-objective functions. This is why Caroline modified it.

mboussaa:
Thanks. Good to know.
Another question: is it worth making the "mu" parameter of MOCMAES configurable. It is now hardcoded "mu=2".

I read on internet: 
for mu – The number of parents used to generate the candidates. The higher this number is the better the Parato front coverage will be, but the longer it will take to converge.

What do you think @carolinepacheco @andrewssobral

andrewssobral:
> Thanks. Good to know. Another question: is it worth making the "mu" parameter of MOCMAES configurable. It is now hardcoded "mu=2".
> 
> I read on internet: for mu – The number of parents used to generate the candidates. The higher this number is the better the Parato front coverage will be, but the longer it will take to converge.
> 
> What do you think @carolinepacheco @andrewssobral

Yes, I agree with you @mboussaa , but let's do it in another PR because it's a new feature, we should create some conditional workflow variables to address the internal parameters of each algorithm.

andrewssobral (approved):

PR #1138, opened 2022-05-12
Update of package json files for activeeon hub

mbenguig:


aminelouati (approved):
A small comment: why CloudAutomation and not Cloud Automation ?

mbenguig:
> A small comment: why CloudAutomation and not Cloud Automation ?

Indeed I leave it like this in the clearwater json file. But we are supposed to bring a major update to all tags/name/slugs in a close future (and maybe remove some projects incl. clearwater)

PR #1170, opened 2022-07-27
Bug fix: Remove single quotes from datetime descriptions

mboussaa:
single quotes seem to cause issues when added to the description of variables. encoding it as html: & # 3 9 ; does not help either. We should not use single quotes in variable descriptions anymore

### DIFF
--- a/Cron/resources/catalog/start_at.xml
+++ b/Cron/resources/catalog/start_at.xml
@@ -3,7 +3,7 @@
         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xmlns="urn:proactive:jobdescriptor:3.13" xsi:schemaLocation="urn:proactive:jobdescriptor:3.13 http://www.activeeon.com/public_content/schemas/proactive/jobdescriptor/3.13/schedulerjob.xsd"  name="Start_At" projectName="1. Basic Workflows" priority="normal" onTaskError="continueJobExecution"  maxNumberOfExecution="2"  >
   <variables>
-    <variable name="DATETIME" value="2019-01-01T12:30:00+02:00" model="PA:DATETIME(yyyy-MM-dd&#39;T&#39;HH:mm:ssXXX)" description="Date and time given in format &lt;b&gt;yyyy-MM-dd&#39;T&#39;HH:mm:ssXXX&lt;/b&gt;&lt;br/&gt; Example:&lt;br/&gt; 2021-10-24T12:54:00+02:00"   />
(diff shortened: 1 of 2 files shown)


Cron/resources/catalog/start_at.xml
```diff
@@ -3,7 +3,7 @@
         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xmlns="urn:proactive:jobdescriptor:3.13" xsi:schemaLocation="urn:proactive:jobdescriptor:3.13 http://www.activeeon.com/public_content/schemas/proactive/jobdescriptor/3.13/schedulerjob.xsd"  name="Start_At" projectName="1. Basic Workflows" priority="normal" onTaskError="continueJobExecution"  maxNumberOfExecution="2"  >
   <variables>
-    <variable name="DATETIME" value="2019-01-01T12:30:00+02:00" model="PA:DATETIME(yyyy-MM-dd&#39;T&#39;HH:mm:ssXXX)" description="Date and time given in format &lt;b&gt;yyyy-MM-dd&#39;T&#39;HH:mm:ssXXX&lt;/b&gt;&lt;br/&gt; Example:&lt;br/&gt; 2021-10-24T12:54:00+02:00"   />
+    <variable name="DATETIME" value="2019-01-01T12:30:00+02:00" model="PA:DATETIME(yyyy-MM-dd&#39;T&#39;HH:mm:ssXXX)" description="Date and time given in format &lt;b&gt;yyyy-MM-ddTHH:mm:ssXXX&lt;/b&gt;&lt;br/&gt; Example:&lt;br/&gt; 2021-10-24T12:54:00+02:00"   />
```
fviale, line 6: the problem is that the format does contain a singlequote

fviale: Ah, but it's ok, the format is 2019-01-01T12:30:00+02:00 so indeed the description should not contain a single quote

fviale (approved):

Cron/resources/catalog/start_at.xml
mboussaa: @fviale 

The format yes and it is ok, there is no issue, but in the description we don't need the single quote because the final result should be like the example 2021-10-24T12:54:00+02:00

no single quotes are required on the T 

I would rather say that the date format is wrong

fviale: The singlequote is necessary in the format as it is required by the date format pattern syntax:
https://docs.oracle.com/javase/8/docs/api/java/text/SimpleDateFormat.html (see examples)

Ok to remove it in the description.

PR #1179, opened 2022-08-25
Update storm to fix freez in while loop and add node selection with hostname

rsaidane:


Storm/resources/catalog/Storm.xml
```diff
@@ -47,21 +60,27 @@ docker pull storm:1.2.2
       <post>
```
fviale, line 60: the post script is empty, and should be deleted

fviale (approved):
LGTM, minor comment

PR #1183, opened 2022-09-02
Replace java docker image by adoptopenjdk/openjdk8

mboussaa:


fviale:
you need to change as well the value of the DOCKER_IMAGE variable in docker_exec_env.xml 
`<variable name="DOCKER_IMAGE" value="java" ...` 

Please check if there are other similar variable set to java

mboussaa:
> you need to change as well the value of the DOCKER_IMAGE variable in docker_exec_env.xml `<variable name="DOCKER_IMAGE" value="java" ...`
> 
> Please check if there are other similar variable set to java

ok done. no other occurences

fviale (approved):

aminelouati (approved):

PR #1210, opened 2022-12-09
Improve Service_Start script to take a PA:CATALOG_OBJECT parameter and not only a SERVICE_ID to identify a PSA service

mklkun:


PcaScripts/resources/catalog/Service_Start.groovy
```diff
@@ -36,11 +63,48 @@ if (binding.variables["args"]) {
     }
 }
 
-
-//Check that the provided serviceId belongs to the existing Service Activation list
-def catalogRestApi = new CatalogRestApi(apiClient)
-if(!catalogRestApi.listAllWorkflowsByServiceIdUsingGET(sessionId).keySet().contains(serviceId)){
-    throw new IllegalArgumentException("The provided SERVICE_ID:<" + serviceId + "> does not belong to the existing Service Activation list. You have to specify an existing service id.")
+if (!Strings.isNullOrEmpty(serviceId)) {
```
fviale, line 66: are you sure that this code should be executed when serviceId is not empty ?

PcaScripts/resources/catalog/Service_Start.groovy
```diff
@@ -36,11 +63,48 @@ if (binding.variables["args"]) {
...
+                bucketName = listStartingWorkflowsByServiceId.get(serviceIdIterator)[0].getBucket()
+                serviceVariables = listStartingWorkflowsByServiceId.get(serviceIdIterator)[0].getVariables().collectEntries {var -> [var.getName(), var.getValue()]}
+                break
+            }
+        }
+    }
+} else if (serviceActivationWorkflow != null) {
+    //Identifying the starting workflow, the service ID and the default variables inside the catalog
+    def cloudAutomationWorkflow
+    Map<String, List<CloudAutomationWorkflow>> listStartingWorkflowsByServiceId = catalogRestApi.listStartingWorkflowsByServiceIdUsingGET(sessionId)
```
fviale, line 96: from what I understand this returns the list of all service automation workflows. This is a SLOW request.

I really think it would be much more efficient from the backend point of view to return only the information needed.

mklkun: Is it slow because of the amount of data sent back or the backend calculation?
Because from a backend point of view it is not since we are using a cache mechanism to fasten it and the list of services is not changing a very often.

fviale: yes, because of the amount of data

mklkun: Fair enough. I will improve that.

PcaScripts/resources/catalog/Service_Start.groovy
```diff
@@ -36,11 +63,41 @@ if (binding.variables["args"]) {
...
+    // Get the service workflow
+    // Check that the provided serviceId belongs to the existing Service Activation list
+    List<CloudAutomationWorkflow> listStartingWorkflows = catalogRestApi.getWorkflowsByServiceIdUsingGET(sessionId, serviceId)
+    if (!listStartingWorkflows) {
+        throw new IllegalArgumentException("The provided SERVICE_ID is not valid.")
+    }
+    // Check when SERVICE_ID and SERVICE_ACTIVATION_WORKFLOW are provided the specific service activation workflow
+    if (serviceActivationWorkflow != null) {
+        for (CloudAutomationWorkflow cloudAutomationWorkflow : listStartingWorkflows) {
+            if (cloudAutomationWorkflow.getName().equals(startingWorkflowName) && cloudAutomationWorkflow.getBucket().equals(bucketName)) {
```
fviale, line 76: I don't understand, if you provide a serviceId, do you have to provide a PA:CATALOG_OBJECT also ?

mklkun: The actual PR makes sure we provide only a PA:CATALOG_OBJECT. But for retro-compatibility I check if the serviceId is provided. 
Now, according to Denis, having multiple starting workflows for one service is a feature! 
So, I can imagine a scenario in which a workflow can be configured to be executed ONLY with a specific service (identified by the serviceId). In this case, on submission time we can select which starting workflow to use.
The different scenarios tested:
- only serviceId
- only PA:CATALOG_OBJECT
- Both compatible
- Both not compatible

fviale: ok, got it, thanks for the clarification

fviale (approved):

PR #1218, opened 2022-12-28
Improve Service_Action and Service_Start CATALOG_OBJECT variable models

mklkun:
- Improve Service_Action script to take a PA:CATALOG_OBJECT parameter and not only an ACTION
- Improve Service_Start CATALOG_OBJECT variable models to version 2.0

### DIFF
--- a/BigData/resources/catalog/Docker_Swarm_HDFS_Spark_Service.xml
+++ b/BigData/resources/catalog/Docker_Swarm_HDFS_Spark_Service.xml
@@ -274,7 +274,7 @@ if (result!=null){
     fork="true">
       <variables>
         <variable name="INSTANCE_NAME" value="spark-$PA_JOB_ID" inherited="false" description="The name of the service to be deployed"/>
-        <variable name="ACTION" value="Finish_Spark" inherited="false" description="The action that will be processed regarding the service status."/>
+        <variable name="SERVICE_ACTION_WORKFLOW" value="service-automation/Finish_Spark" inherited="false" model="PA:CATALOG_OBJECT(Workflow/psa,,,%Spark)" description="The action that will be processed regarding the service status." group="" advanced="false" hidden="false"/>
         <variable name="INSTANCE_ID" value="" inherited="false" description="The service instance ID"/>
       </variables>
       <genericInformation>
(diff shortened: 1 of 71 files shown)


BigData/resources/catalog/Docker_Swarm_HDFS_Spark_Service.xml
```diff
@@ -274,7 +274,7 @@ if (result!=null){
     fork="true">
       <variables>
         <variable name="INSTANCE_NAME" value="spark-$PA_JOB_ID" inherited="false" description="The name of the service to be deployed"/>
-        <variable name="ACTION" value="Finish_Spark" inherited="false" description="The action that will be processed regarding the service status."/>
+        <variable name="SERVICE_ACTION_WORKFLOW" value="service-automation/Finish_Spark" inherited="false" model="PA:CATALOG_OBJECT(Workflow/psa,,,%Spark)" description="The action that will be processed regarding the service status." group="" advanced="false" hidden="false"/>
```
AlexFalessi, line 277: It looks like these changes (ACTION variables changes to SERVICE_ACTION_WORKFLOW) are breaking the Service_Action
script execution because of `actionIterator.getName().equals(action)` always beeing false.

You need to do like the Service_Start script and split the value.

mklkun: The test `actionIterator.getName().equals(action)` is checked only if `!Strings.isNullOrEmpty(action)`. This is of course not the case for 13.1.0-SNAPSHOT's version workflows but for other scenarios it can be the case (retrocompatibility is a must for generic scripts!).

AlexFalessi: Sorry didn't saw your update on the Service_Action script... Disregard my previous comment please

PcaScripts/resources/catalog/Service_Action.groovy
```diff
@@ -29,26 +30,41 @@ if (!instanceId && !instanceName){
...
+            isActionExists = true
+            break
+        }
+    }
+} else if (!Strings.isNullOrEmpty(serviceActionWorkflow)) {
+    def serviceActionWorkflowSplits = serviceActionWorkflow.split('/')
+    bucketName = serviceActionWorkflowSplits[0]
+    action = serviceActionWorkflowSplits[1]
+    CloudAutomationWorkflow executableAction = catalogRestApi.getExecutableActionByCatalogObjectUsingGET(sessionId, instanceId, bucketName, action)
+    println("Action Bucket_name: " + bucketName + ", Action_workflow_name: " + action)
```
AlexFalessi, line 61: You want to keep this print ?

AlexFalessi (approved):

PcaScripts/resources/catalog/Service_Action.groovy
mklkun: Yes. This print is present only if the `SERVICE_ACTION_WORKFLOW` is provided. It is a way (from outputs) to detect if the variables are set in the right way.

mboussaa (approved):

PR #1256, opened 2023-04-11
Adding some AWS Lambda function templates

aminelouati:


ApplicationConnectors/resources/catalog/AWS_Lambda_Function.xml
```diff
@@ -0,0 +1,272 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<job
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xmlns="urn:proactive:jobdescriptor:3.14" xsi:schemaLocation="urn:proactive:jobdescriptor:3.14 http://www.activeeon.com/public_content/schemas/proactive/jobdescriptor/3.14/schedulerjob.xsd"  name="AWS_Lambda_Function" projectName="Cloud Services"  priority="normal" tags="Lambda,Serverless,AWS,Cloud" onTaskError="continueJobExecution"  maxNumberOfExecution="2"  >
+  <variables>
+    <variable name="USERNAME" value="my_user" model="PA:NOT_EMPTY_STRING" description="Specify the username of the system account under which the command will run" group="User Authentication" advanced="true" hidden="false"/>
+    <variable name="PASSWORD" value="my_pwd" model="PA:NOT_EMPTY_STRING" description="Specify the username of the system account under which the command will run" group="User Authentication" advanced="true" hidden="false"/>
```
AlexFalessi, line 7: Shouldn't it be HIDDEN type for the password ?

Also password instead of username in the description.

ApplicationConnectors/resources/catalog/AWS_Lambda_Function.xml
```diff
@@ -0,0 +1,272 @@
...
+    <variable name="PROFILE" value="default" model="" description="Specify the profile to use from the credential file." group="Lambda Authentication" advanced="false" hidden="false"/>
+    <variable name="REGION" value="eu-west-3" model="" description="Specify the region to use in your command." group="Lambda Authentication" advanced="false" hidden="false"/>
```
AlexFalessi, line 9: Can't the model be NOT_EMPTY_STRING for these variables ?

ApplicationConnectors/resources/catalog/AWS_Lambda_Function.xml
aminelouati: The HIDDEN model is limited when this variable is referenced in the GI

ApplicationConnectors/resources/catalog/AWS_Lambda_Function.xml
aminelouati: In case these variables are empty, the default profile and region will be used. I agree that to the description need to be improved.

AlexFalessi (approved):
Small non blocking comments

ApplicationConnectors/resources/catalog/AWS_Lambda_Function.xml
fviale: @aminelouati this is why RUNAS supports 3rd-party credentials. Also the description of the password variable is invalid

fviale:
@aminelouati I will need a general explanation as how the workflow works

ApplicationConnectors/resources/catalog/AWS_Lambda_Function.xml
```diff
@@ -0,0 +1,272 @@
...
+OUTFILE="$variables_OUTFILE"
+
+
+# Check whether PROFILE and REGION have been well entered
+if [ \( ! -z "$PROFILE" -a -z "$REGION" \) -o \( -z "$PROFILE" -a ! -z "$REGION" \) ]; then
+    echo ERROR: "$variables_PA_JOB_NAME"_PROFILE and "$variables_PA_JOB_NAME"_REGION are used in junction. They should be either both entered or both blank.
+    exit 1
+fi
+
+if [ "$EXECUTION_MODE" == "native" ]; then
```
fviale, line 66: what is the benefit of switching between native and docker execution ? Using docker only would allow to use a fork env docker that would make this workflow portable.

aminelouati: The benefit is to have both execution modes. The native mode is the default mode as the customer may not use Docker. The containerized mode is useful for those who do not have preliminary installation of the AWS CLI tools.

ApplicationConnectors/resources/catalog/AWS_Lambda_Function.xml
aminelouati: OK Thanks

PR #1305, opened 2023-09-25
Add Kubernetes-AWS workflows

nebilbenmabrouk:
Two workflows are added: 

1. **Kops_Install_K8S_In_AWS**, which uses the Kops tool to install an enterprise-grade Kubernetes cluster in AWS.
2. **Create_Kops_User_In_AWS**, which creates a user and a group in AWS. The credentials of this user are passed as input in the first workflow (Kops_Install_K8S_In_AWS)


alijawadfahs (approved):

InfrastructureAsCode/resources/catalog/Create-Kubernetes-Cluster-With-Kops.sh
```diff
@@ -0,0 +1,30 @@
...
+
+# Display environment variables before starting the cluster creation
+printenv
+
+# Execute kops commands to create a Kubernetes cluster
+kops create cluster  \
+    --name=${NAME} \
+    --cloud=aws \
+    --zones=${ZONES} \
+    --node-count=${NODE_COUNT} \
```
mboussaa, line 11: why not having a dynamic cluster nodes ? with min and max number of nodes? @nebilbenmabrouk

InfrastructureAsCode/resources/catalog/Create-Kubernetes-Cluster-With-Kops.sh
```diff
@@ -0,0 +1,30 @@
...
+    --node-size=${NODE_SIZE} \
+    --discovery-store=s3://${OIDC_BUCKET}/${NAME}/discovery \
+    --ssh-public-key=/root/.ssh/id_rsa.pub \
+    --cloud-labels="cluster=${NAME}" \
+    --yes 
+#    --dry-run --output yaml \
+    
+#kops update cluster --name ${NAME} --yes --admin
```
mboussaa, line 19: to remove comment

InfrastructureAsCode/resources/catalog/Create-Kops-User-In-AWS.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+
```
mboussaa, line 2: I don't understand why do you need a user to run kops. And it is an AWS user, not KOPS user. You should clarify this somewhere.
Also, define and explain why you need a bucket creation.
Clarify in the description?

InfrastructureAsCode/resources/catalog/Create-Kubernetes-Cluster-With-Kops.sh
nebilbenmabrouk: During the creation of a Kubernetes cluster, Kops supports only the option _--node-count_ which is set to 1 by default. There is no min and max node count at the creation time. Once the cluster is created, the admin can edit the configuration of the cluster using commands, such as:
- kops edit ig (to add min and max node counts to an existing instance group), 
- kops create ig morenodes  (to create a new instance group), 
- kops toolbx instance-selector (to create a new instance group via toolbox)

InfrastructureAsCode/resources/catalog/Create-Kubernetes-Cluster-With-Kops.sh
nebilbenmabrouk: OK

InfrastructureAsCode/resources/catalog/Create-Kops-User-In-AWS.sh
nebilbenmabrouk: - Kops requires a user in AWS IAM (with specific privileges) in order to be able to create all the resources required by the cluster.
- Kops requires 2 buckets in AWS S3 (with specific privileges) in order to be able to store the data of the cluster.
- All these aspects are explained in Kops deocumentation (https://kops.sigs.k8s.io/getting_started/aws/). I added the link to the workflow.

PR #1335, opened 2023-12-01
Improve automatic catalog objects staging

aminelouati:


nebilbenmabrouk (approved):
LGTM if you decided to keep the workflows in basic-examples.

PR #1352, opened 2024-01-25
Update Web_Validation and Email_Validation following new authorized users and user groups feature + Client generation

AlexFalessi:


AlexFalessi:
Replaced by https://github.com/ow2-proactive/proactive-examples/pull/1358

PR #1360, opened 2024-02-21
Fix method name and parameter calls

andrewssobral:


AlexFalessi (approved):
OK approved
. 
`launchServiceInstanceAction()` is called at multiple places and we need to update the all.
Do you want to add a commit here or do you want me to do it in a different PR ?

PR #1364, opened 2024-02-22
Use 'as Long' instead of Long.parseLong() as we don't know the type o…

AlexFalessi:
…f the INSTANCE_ID variable here


fviale (approved):

nebilbenmabrouk (approved):
LGTM

PR #1367, opened 2024-03-07
Add more workflows tests to System Tests + Add necessaray SKIP_SIGNAL…

AlexFalessi:
… to bypass waitin for a signal to terminate job and services + Fix potential parseLong exception

### DIFF
--- a/AutoML/resources/catalog/Distributed_Auto_ML.xml
+++ b/AutoML/resources/catalog/Distributed_Auto_ML.xml
@@ -2065,7 +2065,7 @@ if (TENSORBOARD_ENABLED) {
         if( !actionVariables.isEmpty() ){
             actionVariables.each{ k, v -> service.putVariablesItem("${k}", "${v}") }
         }
-        def serviceInstanceData = serviceInstanceRestApi.launchServiceInstanceAction(sessionId, Long.parseLong(instanceId), service, Integer.parseInt(variables.get("PA_JOB_ID")))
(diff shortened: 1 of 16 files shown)


AutoML/resources/catalog/Distributed_Auto_ML.xml
```diff
@@ -2065,7 +2065,7 @@ if (TENSORBOARD_ENABLED) {
         if( !actionVariables.isEmpty() ){
             actionVariables.each{ k, v -> service.putVariablesItem("${k}", "${v}") }
         }
-        def serviceInstanceData = serviceInstanceRestApi.launchServiceInstanceAction(sessionId, Long.parseLong(instanceId), service, Integer.parseInt(variables.get("PA_JOB_ID")))
+        def serviceInstanceData = serviceInstanceRestApi.launchServiceInstanceAction(sessionId, instanceId as Long, service, Integer.parseInt(variables.get("PA_JOB_ID")))
```
aminelouati, line 2068: Are you sure from this Long casting? (instead of using Long.parseLong())

AlexFalessi: Yes It's better use `as Long` because `parseLong()` accepts only a String as param where here it will work also with integer, String or even long.
Before we did `as int`, which i first change to `parseLong()` because of client regeneration.

aminelouati (approved):

AlexFalessi:
retest this please

PR #1386, opened 2024-05-02
Changes on MLOps_Model_Server_Application_Inference and MLOps_Drift_Notification_Workflow workflows

dianajlailaty:
Changes of this PR:
For the **MLOps_Model_Server_Application_Inference** workflow:

1. Adding iris-classification-model into the list of models to use for sending inferences.
2. Adding DRIFT_DATA variables. This variable only appears if iris-classification-model is chosen. If True, input data will be drifted by adding noise (using Caroline's method).
3. Adding a Simulate_Data_Drift Signal. It allows the user to change the DRIFT_DATA value. However, it obliges the user to choose iris-classification-model (as DRIFT_DATA is only allowed for iris model). So, if the user had chosen "simple" model and then clicks on Simulate_Data_Drift signal, the model will change from simple to iris-classification-model. For that, I have added a drop list with only the model(s) that allow drift (for now we only have iris-classification-model).

For the **MLOps_Drift_Notification_Workflow**:

1. Changing the timezone mechanism after facing problems with daylight timing (after hour change in March). Now we have a more detailed list of timezones and it takes into consideration the daylight timing condition.


MLOpsModelServer/resources/catalog/MLOps_Drift_Notification_Workflow.xml
```diff
@@ -30,7 +30,7 @@
...
...an/Mayotte, Indian/Reunion, Iran, Israel, Jamaica, Japan, Kwajalein, Libya, MET, MST, MST7MDT, Mexico/BajaNorte, Mexico/BajaSur, Mexico/General, NZ, NZ-CHAT, Navajo, PRC, PST8PDT, Pacific/Apia, Pacific/Auckland, Pacific/Bougainville, Pacific/Chatham, Pacific/Chuuk, Pacific/Easter, Pacific/Efate, Pacific/Enderbury, Pacific/Fakaofo, Pacific/Fiji, Pacific/Funafuti, Pacific/Galapagos, Pacific/Gambier, Pacific/Guadalcanal, Pacific/Guam, Pacific/Honolulu, Pacific/Johnston, Pacific/Kanton, Pacific/Kiritimati, Pacific/Kosrae, Pacific/Kwajalein, Pacific/Majuro, Pacific/Marquesas, Pacific/Midway, Pacific/Nauru, Pacific/Niue, Pacific/Norfolk, Pacific/Noumea, Pacific/Pago_Pago, Pacific/Palau, Pacific/Pitcairn, Pacific/Pohnpei, Pacific/Ponape, Pacific/Port_Moresby, Pacific/Rarotonga, Pacific/Saipan, Pacific/Samoa, Pacific/Tahiti, Pacific/Tarawa, Pacific/Tongatapu, Pacific/Truk, Pacific/Wake, Pacific/Wallis, Pacific/Yap, Poland, Portugal, ROC, ROK, Singapore, Turkey, UCT, US/Alaska, US/Aleutian, US/Arizona, US/Central, US/East-Indiana, US/Eastern, US/Hawaii, US/Indiana-Starke, US/Michigan, US/Mountain, US/Pacific, US/Samoa, UTC, Universal, W-SU, WET, Zulu)" name="TIMEZONE" value="Europe/Paris"/>
```
andrewssobral, line 33: @dianajlailaty we may should consider this into the MLOps Dashboard too.

andrewssobral (approved):
Approved after refactoring

PR #1390, opened 2024-06-25
Improve file data connectors and event file listener

aminelouati:


WebNotification/resources/catalog/Web_Notification_Script.groovy
```diff
@@ -4,6 +4,12 @@ import org.ow2.proactive.notification.client.model.EventRequest
 import org.ow2.proactive.notification.client.model.Event
 import org.ow2.proactive.notification.client.ApiException
 
+// Skip web notification
+if (variables.get("ACTIVATE_NOTIFICATIONS") && variables.get("ACTIVATE_NOTIFICATIONS").toBoolean()) {
```
AlexFalessi, line 8: I think you should negate here = `!variables.get("ACTIVATE_NOTIFICATIONS").toBoolean()`
if ACTIVATE_NOTIFICATIONS is false we skip. Or change the var name

AlexFalessi (approved):
Approved with small comment

WebNotification/resources/catalog/Web_Notification_Script.groovy
aminelouati: you are right ! Thanks

PR #1394, opened 2024-07-24
Enable the access to knime service HTTP interface via the IP address

aminelouati:


Knime/resources/catalog/Knime.xml
```diff
@@ -131,6 +135,44 @@ import org.ow2.proactive.pca.service.client.model.Container
...
+    } catch (IOException e) {
+        e.printStackTrace()
+    }
+    // Return null if an error occurs
+    return null
+}
+
+// Regex expression for validating IPv4
+String regex="(([0-9]|[1-9][0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5])\\.){3}([0-9]|[1-9][0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5])"
+Pattern p = Pattern.compile(regex)
```
fviale, line 174: p is a bit dangerous for a variable name as it could be used somewhere else without you knowing, could you call it ipPattern ?

fviale (approved):

PR #1397, opened 2024-08-22
fix web validation related workflows

aminelouati:


WebNotification/resources/catalog/Web_Validation_Script.groovy
```diff
@@ -1,5 +1,6 @@
 // Skip web validation (used for automatic workflow tests to skip manual validation)
 if (variables.get("SKIP_WEB_VALIDATION") && variables.get("SKIP_WEB_VALIDATION").toBoolean()) {
+    sleep(10000)
```
AlexFalessi, line 3: This was to "simulate" the time needed by the user to process the validation.
You can remove it if you want.

AlexFalessi (approved):

WebNotification/resources/catalog/Web_Validation_Script.groovy
aminelouati: This sleep is intentionally added to allow the PSA service to transit from one state to another before executing the following task in the workflow. Otherwise, the job fails

PR #1409, opened 2024-10-23
Add send Slack message workflow

AlexFalessi:
Add a workflow to send a message on aslack channel using a webhook

### DIFF
--- a/Slack/METADATA.json
+++ b/Slack/METADATA.json
@@ -1,0 +1,24 @@
+{
+  "metadata": {
+    "slug": "slack",
+    "name": "Slack",
+    "short_description": "Workflows to interact with a Slack workspace",
+    "author": "ActiveEon's Team",
+    "version": "1.0"
+  },
+  "catalog": {
+    "bucket": "control-notification",
+    "userGroup" : "public-objects",
+    "objects": [
+      {
+        "name": "Slack_Message",
+        "metadata": {
+          "kind": "Workflow/standard",
+          "commitMessage": "First commit",
+          "contentType": "application/xml"
+        },
+        "file": "resources/catalog/Slack_Message.xml"
+      }
+    ]
+  }
+}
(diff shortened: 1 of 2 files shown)


Slack/resources/catalog/Slack_Message.xml
```diff
@@ -0,0 +1,125 @@
...
+  <taskFlow>
+    <task fork="true" name="Send_Message">
+      <description>
+        <![CDATA[ Executes a HTTP Post request to a slack webhook. ]]>
+      </description>
+      <variables>
+        <variable advanced="false" description="" group="" hidden="false" inherited="false" name="WEBHOOK" value="https://hooks.slack.com"/>
+        <variable advanced="false" description="" group="" hidden="false" inherited="false" name="MESSAGE_HEADER" value=""/>
+        <variable advanced="false" description="" group="" hidden="false" inherited="false" name="MESSAGE_CONTENT" value=""/>
+        <variable advanced="false" description="" group="" hidden="false" inherited="true" name="DEBUG" value="false"/>
```
aminelouati, line 21: Could you please provide a description to all variables please?

Slack/METADATA.json
```diff
@@ -0,0 +1,24 @@
...
+        "metadata": {
+          "kind": "Workflow/standard",
+          "commitMessage": "First commit",
+          "contentType": "application/xml"
+        },
+        "file": "resources/catalog/Slack_Message.xml"
+      }
+    ]
+  }
+}
```
aminelouati, line 24: Could you add a test version into workflow-test ?

fviale:
@AlexFalessi 
According to the webhook documentation, there is another procedure (via an OAuth flow) to generate webhook urls on the fly.

I don't know completely what is the use case here, and if we should consider the second approach

Slack/resources/catalog/Slack_Message.xml
AlexFalessi: Sure I'll add descriptions.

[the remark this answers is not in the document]
AlexFalessi: It in the task variables as inherited. You're right I'll put it as advanced.

AlexFalessi:
@fviale 
It looks like generating webhooks on the fly using the OAuth flow is part of the application installation process where a user installs an app, and in the authentication response a webhook is provided (generated on the fly).
Instead of webhooks defined in the app itself.

Here we just want a webhook with no focus on how to get it.

We don't have an ActiveEon slack app, we use the deprecated https://slack.com/marketplace/A0F7XDUAZ-incoming-webhooks
for our custom webhooks.

An AE slack app would be a good idea and enable a lot of features and complex flows between AE and Slack

Slack/METADATA.json
AlexFalessi: To execute it in the test requires to effectively post a slack message in order to have a 200 response.
Otherwise the wokflow fails with an exception.

Should we still execute it ?

AlexFalessi: I added a SKIP_FAILURE variable check to skip the exception

